PR #98, opened 2021-08-18
feat: Add user agent parsing

### LINKED ISSUE #66: Add user agent parsing
We need to parse the user-agent header of requests to derive information we need to send to AdM. This information should become a part of the `SuggestionRequest` object, and be included in cache keys. The data we nee to derive from the user-agent are:

* `form-factor` desktop, tablet, mobile, or other
* `os-family` windows, macOS, linux, iOS, android, chromeOS, blackBerry, other

Of course some of these os-family values are less likely than others. :grin: 

See the Firefox Conducive API spec from adM for more details.

**Acceptance Criteria:**

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1131)
┆Epic: merino-m0

ethowitz:
Closes #66

### DIFF
--- a/merino-suggest/src/lib.rs
+++ b/merino-suggest/src/lib.rs
@@ -471,7 +471,7 @@ pub struct DeviceInfo {
     pub os_family: OsFamily,
     /// The device form factor indicated in the User-Agent header.
     pub form_factor: FormFactor,
-    /// The major browser version of Firefox .
+    /// The web browser indicated in the User-Agent header.
     pub browser: Browser,
 }
 
--- a/merino-suggest/src/lib.rs
+++ b/merino-suggest/src/lib.rs
@@ -29,6 +29,9 @@ pub use crate::debug::DebugProvider;
 pub use crate::multi::Multi;
 pub use crate::wikifruit::WikiFruit;
 
+/// The range of major Firefox version numbers to use for testing.
+pub const FIREFOX_VERSION_RANGE: Range<u32> = 70..95;
+
 /// A request for suggestions.
 #[derive(Debug, Clone, Hash, Serialize)]
 pub struct SuggestionRequest<'a> {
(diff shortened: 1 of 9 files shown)


merino-suggest/src/lib.rs
```diff
@@ -341,6 +350,151 @@ pub enum LanguageIdentifier {
     Wildcard,
 }
 
+/// The form factor of the device that sent a given suggestion request.
+#[derive(Clone, Debug, Hash, PartialEq, Serialize)]
```
pjenvey, line 354: Let's move these new structs to a separate file, this module's getting pretty large. I'm big about doing this up front rather than later because we'll be less likely to run into merge conflicts with each other in the future, especially during a crunch time (which hopefully will not occur in the first place :smile: ).

Plus it's always more difficult to do later for the same reason as the merge conflicts.

jrconlin: not to muddy things, but kinda/sorta wonder if this might be a good candidate for `common`, since it's basically the same code being used in both  Merino and Contile.

Counter argument is that this code is also very specific to Contile and Merino and it's fine having the duplication since we're "fuzzing" things.

mythmon: Let's land this PR approximately as is. I agree that adding something like this to common would be a good idea in the future.

merino/src/docs/api.rs
```diff
@@ -38,6 +38,9 @@ and headers.
   Merino will return an empty list instead of attempting to query other
   languages.
 
+- `User-Agent` - A user's device form factor, operating system, and browser/Firefox version
```
ethowitz, line 41: I wasn't sure what else to say here (since we don't actually reject any user agents), but happy to expound further if you have any suggestions!

mythmon: Looks good as it is. Mostly I wanted to make sure that the API noted the use of User Agent. Also I wanted to make sure that we are keeping this doc up to date, so this was good practice.

mythmon (approved):

PR #110, opened 2021-08-24
feat: add locking to redis cache

### LINKED ISSUE #104: Add cache locking to redis cache
When consulting the cache, it will often be the case that multiple requests will be handled between when a cache entry expires and when it can be refreshed. This can create a stampede of re-generating the same content when a cache entry fails, as many workers concurrently update the cached item.

To avoid this problem, cache entries should become locked when they are being regenerated. Clients that encounter a locked cache entry should either serve stale content, an empty request, or wait until the lock has been removed.

This issue covers adding cache locking to the Redis cache.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1230)

jrconlin:
This adds record write locking to redis to prevent multiple threads from
attempting to update a `key` record. This is a VERY simple lock.

Closes #104

### DIFF
--- a/merino-cache/src/redis/mod.rs
+++ b/merino-cache/src/redis/mod.rs
@@ -2,7 +2,7 @@
 
 mod domain;
 
-use std::{borrow::Cow, time::Duration};
+use std::{borrow::Cow, convert::TryInto, time::Duration};
 
 use crate::{domain::CacheKey, redis::domain::RedisSuggestions};
 use anyhow::{anyhow, Context};
@@ -14,6 +14,7 @@ use merino_suggest::{
 };
 use redis::RedisError;
 use tracing_futures::{Instrument, WithSubscriber};
+use uuid::Uuid;
 
 use self::domain::RedisTtl;
 
@@ -28,6 +29,9 @@ pub struct Suggester<S> {
     /// The default amount of time a cache entry is valid, unless overridden by
     /// `inner`.
     default_ttl: Duration,
+
+    /// Default lock timeout
+    default_lock_timeout: Duration,
 }
 
 #[derive(Debug)]
@@ -68,6 +72,7 @@ where
             inner: provider,
             redis_connection,
             default_ttl: settings.redis_cache.default_ttl,
+            default_lock_timeout: settings.redis_cache.default_lock_timeout,
         }))
     }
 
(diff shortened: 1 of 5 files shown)


merino-cache/src/redis/mod.rs
```diff
@@ -125,6 +130,99 @@ where
...
     }
 
+    /// Generate a lock identifier key.
+    /// This is a VERY simple locking mechanism. The only bit of fancy is that it will
+    /// expire, allowing for "stuck" queries to eventually resolve.
+    fn pending_key(key: &str) -> String {
+        format!("pending_{}", key)
+    }
+
+    /// See iff a record update is pending. (Does not check lock value, only if it exists)
```
mythmon, line 140: Is this `iff` a typo? I don't think "if and only if" applies here.

merino-cache/src/redis/mod.rs
```diff
@@ -125,6 +130,99 @@ where
...
+    async fn is_pending(&self, key: &str) -> Result<bool, SuggestError> {
+        let pending_key = Self::pending_key(key);
+        let mut connection = self.redis_connection.clone();
+
+        tracing::trace!("🔒Checking key: {:?}", &pending_key);
```
mythmon, line 145:
```suggestion
        tracing::trace!(%pending_key, "🔒Checking key");
```

You can do string interpolation, but structured logging in Tracing is done separate from that.

merino-cache/src/redis/mod.rs
```diff
@@ -125,6 +130,99 @@ where
...
+        Ok(lock)
+    }
+
+    /// Delete a lock (fails if the lock value does not match.)
+    async fn del_pending(&self, key: &str, lock: &str) -> Result<(), SuggestError> {
+        let mut connection = self.redis_connection.clone();
+        let pending_key = Self::pending_key(key);
+        if self.check_pending(key, lock).await {
+            tracing::trace!("🔒Removing lock for {:?}", &pending_key);
+            redis::cmd("DEL")
```
mythmon, line 212: Instead of having separate check and delete commands (which could be racey), you can use a [`Script`](https://docs.rs/redis/0.21.0/redis/struct.Script.html) command. That lets you send some Lua that will check if the value is the expected UUID, and then delete it, otherwise error out. This is all done automically.

merino-cache/src/redis/mod.rs
jrconlin: yep, typo. Thanks!

merino-cache/src/redis/mod.rs
jrconlin: Ah, cool! TIL.

Now I just have to learn Lua quick.

I'll probably also update the write function to only write if the lock matches as well, because there's a bit of a potential race in there too.

ncloudioj: Yeah, Redis has this [pattern](https://redis.io/commands/set#patterns) to showcase the distributed locking on a single Redis node.

mythmon (changes requested):
Sorry for not mentioning this yesterday (I had to leave before I could properly finish the review), but can you think of a way to test this behavior? I'm struggling to think of something we could with the current infrastructure. Maybe we could put a lock in Redis, and then check that empty results are given until the lock is removed?

Do you know the difference between EVAL and the Redis crate's Script command? I haven't looked too closely at them, so I'm curious if you found a reason to go with one over the other.

Finally, I think we should figure out a policy about the structured data in Tracing logs. I think this is fine to merge as is, but I'd like to get something written down regarding that, and I'd like to hear your thoughts about when you'd use structured logs vs string interpolation.

PR #115, opened 2021-08-25
ci: Fix bad quotes in docker-publish job

mythmon:


jrconlin (approved):
The Audit fail is because of an OpenSSL vuln with a lesser used crypto library (we never use it). We can ignore it.

Will re-approve.

PR #119, opened 2021-09-01
feat(suggest): Add score field to suggestion

### LINKED ISSUE #118: Add score to Merino suggestion responses
In the future, we want to be able to control how clients select suggestions to display from multiple sources. To facilitate that, each suggestion should carry a score value. For now, the Remote Settings version of quick suggest will have a static score of 0.2. Merino can choose scores that relate to that to indicate which suggestion should be preferred.

**Acceptance Criteria**

* Each suggestion in the API response should contain a score field.
* The score field should be in the range of 0.0 to 1.0.
* There should be unit tests covering the change.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1239)

mythmon:
Score indiciates the relative preference between suggestions, with higher
scores being preferred. This can allow Firefox and other clients of Merino to
make more intelligent decisions about what suggestions to include when many are
available.

Fixes #118


ncloudioj (approved):
LGTM. Can you also update the API doc for this?

mythmon:
Oh right. That was sort of the whole point of this. Thanks for the reminder 😅

PR #122, opened 2021-09-07
Set up integration test infrastructure 🤖

### LINKED ISSUE #121: Set up integration test infrastructure for Merino
Add a new job to Merino's CircleCI config for running integration tests against the build for the current revision. The integration test suite runs at least one test that uses a test client to send an HTTP request to a local instance of Merino and verifies that the response is successful.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1245)

hackebrot:
Please note that I updated the `docker-image-build-pr` job in `.circleci/config.yml` to save the Docker image archive to the CircleCI workspace. We will need decide how we want to configure the integration tests job in the workflow before we merge this pull request. 🚧

Resolve #121


.circleci/config.yml
```diff
@@ -31,7 +31,7 @@ workflows:
 
       - docker-image-build:
           name: docker-image-build-pr
-          saveToWorkspace: false
+          saveToWorkspace: true  # Changed to true to test out integration tests job
```
mythmon, line 34: Since this is always true now, I think we can just remove the option to disable it, in the `docker-image-build` section of this file.

mythmon (approved):
This seems reasonable to me, and the tests seem to work.

> We will need decide how we want to configure the integration tests job in the workflow before we merge this pull request.

What kind of integration are you thinking of? I think I'm happy with the current set up, is there something more you wanted to change?

hackebrot:
Hi @mythmon! 👋🏻 Thanks for the review!

> What kind of integration are you thinking of? I think I'm happy with the current set up, is there something more you wanted to change?

No, only whether you're OK to save the Merino Docker image to the CircleCI workspace on pull requests. Related to your inline comment, do you prefer to keep two separate `docker-image-build` jobs with different filters and just remove the `saveToWorkspace` parameter from it or do you prefer to have a single `docker-image-build` without any filters?

mythmon:
The reason I didn't save them to the workspace before was that it took a fair amount of time, and it wasn't used by anything. Now that we have a use for the docker images in PRs, I think we should unify the CI jobs so we don't have unused complexity.

mythmon (approved):

PR #123, opened 2021-09-07
feat: Add memory cache locking to prevent spurious update requests.

jrconlin:
This uses a simple HashMap wrapped by [arc_swap] to prevent collisions and excessive write locks. Not super happy with some of this, but should work ok.

TODO:

- [ ] add tests
- [ ] cleanup, fmt, clippy

Closes: #105


merino-cache/src/memory.rs
```diff
@@ -16,13 +16,59 @@ use merino_suggest::{
...
 use tracing::Instrument;
 
+use arc_swap::ArcSwap;
+
+lazy_static! {
+    static ref LOCK_TABLE: ArcSwap<HashMap<String, Instant>> =
+        ArcSwap::from_pointee(HashMap::new());
+}
+
+impl LOCK_TABLE {
```
mythmon, line 34: Oh, that's really interesting. I didn't realize you could do that with `lazy_static`.

merino-cache/src/memory.rs
```diff
@@ -70,7 +120,7 @@ impl<S> Suggester<S> {
     fn remove_expired_entries<K: Eq + Hash + Debug>(
         items: &Arc<DedupedMap<K, Instant, Vec<Suggestion>>>,
     ) {
-        let start = Instant::now();
+        let mut start = Instant::now();
```
mythmon, line 123: Does this need to be mutable?

merino-cache/src/memory.rs
```diff
@@ -16,13 +16,59 @@ use merino_suggest::{
...
+        self.rcu(|table| {
+            let mut locked = HashMap::clone(table);
+            locked.insert(key.to_owned(), lock);
+            locked
+        });
+        lock
+    }
+
+    /// run func only if the lock we have matches what is registered for the key
+    fn update<F>(&self, key: &str, lock: Instant, mut func: F)
```
mythmon, line 55: This function also unlocks the lock after the update happens right? That sounds like something that should be mentioned in the doc string or the function name.

mythmon (changes requested):
This looks tricky. I'm not 100% sure about the details of `arc_swap`, but it seems like it should work. A couple much smaller changes, and I think this should be good to go.

merino-cache/src/memory.rs
```diff
@@ -93,6 +143,13 @@ impl<S> Suggester<S> {
             ControlFlow::Continue(!should_remove)
         });
 
+        // remove any expired elements from the Pending table (There shouldn't be many.)
+        LOCK_TABLE.rcu(|table| {
+            let mut cleaned = HashMap::clone(table);
+            cleaned.retain(|_k, v| v > &mut start);
```
mythmon, line 149:
```suggestion
            cleaned.retain(|_k, v| *v > start);
```

merino-cache/src/memory.rs
jrconlin: It is a bit of a cheat, but rust allows you to `impl` a surprising number of things.

merino-cache/src/memory.rs
jrconlin: Probably not, for the same fix that you note later.

merino-cache/src/memory.rs
jrconlin: Fixing this, but also moving it to it's own function in the `LOCK_TABLE` impl, because it probably should live there.

merino-cache/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2018"
 
 [dependencies]
 anyhow = "^1"
-arc-swap = "1.3"
```
mythmon: Is there something in 1.3.2 that we need?

mythmon (approved):

mythmon:
I rebased and force-pushed this to pick up CI changes from `main`.

merino-cache/Cargo.toml
jrconlin: No, this was in part due to the rust version update and making sure we got a good compile. We can probably relax that fairly quickly. If you like, I can add a comment.

mythmon: No, I think it's fine as is. I just wanted to make sure I wasn't missing something in their changelog.

PR #126, opened 2021-09-10
feat: add variant fields to suggest api

### LINKED ISSUE #96: Add variant fields to suggest API to support experimentation
To support future experimentation, the API should accept a `client_variants` field and send both a `server_variants` field and `client_variants` field. Each of these should be a list of strings. For now these won't do anything, but we should add them to the API documentation, and report metrics for received client variants.

**Acceptance criteria**

* Any contents of a `client_variants` query parameter gets parsed and returned in a `client_variants` field at the top level of the suggestion response.
* There is an empty `server_variants` field in the top level of the suggestion response.
* Variants parsed from `client_variants` should emit count metrics.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1213)

tiftran:
fixes #96


jrconlin (approved):

jrconlin:
@tiftran Not sure why those two steps have hung. I wonder if "jiggling the handle" might be needed. You can either try force pushing the PR again, or you can slightly alter a file to trigger a new commit and send that through. 

It happens sometimes.

mythmon:
Yes, the two "hung" tests are reflections of the newer CI configuration that is on `main`. Merging that in or rebasing this on `main` will cause this branch to run the expected jobs.

hackebrot (commented):
Hi @tiftran! 👋🏻 

The CircleCI workflow experienced a Docker issue which caused the integration test job to fail. I re-ran the workflow from the failing job which worked.

The integration tests now failed with an actual error (see [CircleCI log](https://app.circleci.com/pipelines/github/mozilla-services/merino/291/workflows/008645c1-b3b9-42c2-bcca-cd92a015ce95/jobs/1211?invite=true#step-105-313)):

```text
assert r.json() == step.response.content.dict()
Left contains 2 more items:
{'client_variants': [], 'server_variants': []}
```

I want to describe the steps for resolving this below:

1. Following the "Process for investigating failures" from our "Guidelines for disabling automated tests" document, we need to check whether the business requirements have changed
2. You've implemented a new feature for Merino which added two new fields to the `SuggestResponse`. The integration tests did not expect these new fields which caused the above error
3. We can now safely update the tests, because we know that the business requirements have changed and the tests are in fact not broken

The integration test framework defines a model for the response from Merino in `test-engineering/client/tests/models.py`.

```python
class Suggestions(BaseModel):
    """Class that contains a list of Suggestions returned by Merino."""

    suggestions: List[Suggestion]


class Response(BaseModel):
    """Class that holds information about a HTTP response from Merino."""

    status_code: int
    content: Union[Suggestions, Any]
    headers: List[Header] = []
```

The first step is to add `client_variants` and `server_variants` to the above [pydantic model](https://pydantic-docs.helpmanual.io/usage/models/) named `Suggestions`. Then we need to update the expected responses for the integration tests in `test-engineering/volumes/client/scenarios.yml` to contain empty JSON arrays for both fields. That should make the integration tests pass.

My recommendation would be to also add a new integration test which sends a `client_variants` query parameter in the request, so we have coverage for this feature in the client facing interface. @ashrivastava-qa, can you create a JIRA ticket for this and work with @tiftran on that? 📋

Hope this helps. Please let me know if you have any questions!

test-engineering/client/tests/models.py
```diff
@@ -42,6 +42,8 @@ class Suggestions(BaseModel):
     """Class that contains a list of Suggestions returned by Merino."""
 
     suggestions: List[Suggestion]
+    client_variants: List[str]
```
hackebrot, line 45: I think we need `pydantic` to generate the default value for us, so that the response model always contains an empty list. We need to import `Field` and then do the following:
```suggestion
    client_variants: List[str] = Field(default_factory=list)
```

test-engineering/client/tests/models.py
```diff
@@ -42,6 +42,8 @@ class Suggestions(BaseModel):
...
+    server_variants: List[str]
```
hackebrot, line 46:
```suggestion
    server_variants: List[str] = Field(default_factory=list)
```

hackebrot:
Hi @tiftran! 👋🏻 

There seemed to be an issue with CircleCI not building the `client` Docker image even though the source code for it changed. I have added the below step before we run `docker-compose up` to `.circleci/config.yml` to make sure CI always builds the Docker image for the test client before we run the integration tests:

```text
docker-compose -f test-engineering/docker-compose.yml build client
```

We pull `mozilla/contile-integration-tests-partner` and `mozilla/contile-integration-tests-client` from DockerHub for Contile, so this is where our setup for Merino is different. My apologies for missing that. 🙁 

CI now passes for your pull-request including the integration you've added for the `client_variants` query parameter.

mythmon:
@tiftran seems like this is all ready now. Is there anything else you wanted to do before we take it out of draft?

PR #129, opened 2021-09-16
feat: Add setting to configure sentry env

### LINKED ISSUE #124: Add setting to configure Sentry environment
When we added Sentry integration to Merino, it used the value of `MERINO_ENVIRONMENT` for the Sentry environment name. Since Merino's environment is closely tied to settings, it is more about the kind of environment that it is running in, not a specific named environment.

Specifically, Merino's dev, stage, and production servers will all have `MERINO_ENVIRONMENT=production`, so that they use the settings intended for live servers.

However, for Sentry we want to be able to specify which specific environment is being used, and distinguish between dev, stage, and production. We should make Sentry use a dedicated settings instead of using `MERINO_ENVIRONMENT`.

**Acceptance criteria**

* Sentry environment is configurable through a dedicated environment variable
* The setting is documented

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1246)

tiftran:
closes #124

### DIFF
--- a/merino-settings/src/lib.rs
+++ b/merino-settings/src/lib.rs
@@ -222,7 +222,7 @@ pub struct MetricsSettings {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(tag = "mode", rename_all = "snake_case")]
 pub enum SentrySettings {
-    Release { dsn: Dsn },
+    Release { dsn: Dsn, env: String },
     Debug,
     Disabled,
 }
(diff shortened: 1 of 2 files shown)


hackebrot:
Hi @tiftran! 👋🏻 

I restarted the CircleCI workflow as it failed with Docker connectivity issues.

jrconlin (approved):

PR #133, opened 2021-09-20
refactor(adm): Remove usage of remote-settings-client

### LINKED ISSUE #131: Refactor Remote Settings suggestion provider to not use remote-settings-client crate
The [remote-settings-client crate ](https://github.com/mozilla-services/remote-settings-client) isn't targeting a use case like ours, and is missing features we need. Additionally, it complicates our tech stack by referring to older versions of libraries and overall making our lives harder.

Since we're going to be using Remote Settings data more in the future, we should make sure we have a reliable way to consume this data.

**Acceptance Criteria**

* Remove usage of remote-settings-client from Merino
* Use the Remote Settings API to retrieve the quicksuggest collection
* Fetch attachments to build our suggestion library from
* A local disk cache is not neccesary

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1258)
┆Epic: merino-milestone-1

mythmon:
Notably, there is no longer any attempt at all to cache data from Remote
Settings, and signatures are not verified.

Fixes #131

### DIFF
--- a/merino-adm/src/remote_settings.rs
+++ b/merino-adm/src/remote_settings.rs
@@ -3,19 +3,17 @@
 use anyhow::{anyhow, Context};
 use async_trait::async_trait;
 use futures::StreamExt;
-use http::Uri;
+use http::{HeaderValue, Uri};
 use lazy_static::lazy_static;
 use merino_settings::{providers::RemoteSettingsConfig, Settings};
 use merino_suggest::{
     Proportion, SetupError, SuggestError, Suggestion, SuggestionProvider, SuggestionRequest,
     SuggestionResponse,
 };
-use remote_settings_client::client::FileStorage;
-use serde::Deserialize;
-use serde_json::Value;
+use reqwest::Response;
+use serde::{Deserialize, Serialize};
 use serde_with::{serde_as, DisplayFromStr};
 use std::{collections::HashMap, convert::TryFrom, sync::Arc};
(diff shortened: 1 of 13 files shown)


merino-adm/src/remote_settings.rs
```diff
@@ -60,53 +53,47 @@ impl RemoteSettingsSuggester {
...
-            .into_iter()
-            .filter(|r| !r.deleted())
-            .map(|r| {
-                let value = Value::Object(r.as_object().clone());
-                <SuggestRecord as Deserialize>::deserialize(value)
-            })
-            .collect::<Result<_, <Value as serde::Deserializer>::Error>>()
-            .context("Parsing suggestions records")
-            .map_err(SetupError::Format)?;
+                next_url = match records_res.headers().get("Next").map(|v| v.to_str()) {
```
pjenvey, line 79: Is this raw kinto being hit here?

I don't any references to a "Next" header but I see [a "Next-Page"](https://docs.kinto-storage.org/en/latest/api/1.x/pagination.html?highlight=Next-Page)

mythmon: Yes this is hitting Kinto directly. You're right. Dunno how I got just plan `Next`.

pjenvey (approved):

mythmon:
Need to make changes to account for #141 once it merges.

PR #135, opened 2021-09-21
feat(web): Allow specifying providers when requesting suggestions

mythmon:
- refactor(web): Organize files for web endpoints
- dev: Add a provider that suggests static strings for testing
- feat(web): Add endpoint that lists available providers
- feat(web): Allow specifying providers when requesting suggestions


merino/src/docs/api.rs
```diff
@@ -143,4 +153,31 @@ the user opts out the telemetry collection of Firefox.
...
+
+Endpoint: `/api/v1/providers`
+
+This endpoint gives a list of available providers, along with their
+*availability*.  It accepts GET requests and takes no parameters.
+
+### Response
+
+The response will be a JSON object containing the key `providers`, which be a
+map where the keys to this map are the IDs of the provider, and the value are
```
pjenvey, line 166:
```suggestion
The response will be a JSON object containing the key `providers`, which is a
map where the keys to this map are the IDs of the provider, and the values are
```

merino/src/docs/api.rs
```diff
@@ -143,4 +153,31 @@ the user opts out the telemetry collection of Firefox.
...
+provider metadata object. Each provider metadata object will have the following
+format:
+
+- `id` - A string that can be used to identify this provider. This ID can be
+  used for the `providers` field of the suggest API.
+
+- `availability` - A string describing how this provider is used in Merino. It will be one of:
+
+  - `"enabled-by-default"` - This provider will be used for requests that don't
```
pjenvey, line 175: `ProviderAvailability` is serialized as `snake_case` but described as `kebab-case` here (which I just learned serde does support if you want it 🥙)

merino/src/docs/api.rs
```diff
@@ -143,4 +153,31 @@ the user opts out the telemetry collection of Firefox.
...
+    specify providers, and it should be provided to the user as a selection that
+    can be turned off.
+  - `"disabled-by-default"` - This provider is not used automatically. It should
+    be provided to the user as a selection that could be turned on.
+  - `"hidden"` - This provider is not used automatically. It should not be
```
pjenvey, line 180: Is this different from an `is_null()` provider (which would never be included in this list)?

pjenvey (changes requested):

merino/src/docs/api.rs
mythmon: `is_null` is for providers that don't ever return suggestions, and in fact I wanted to do it by something like

```rust
if let Some(_) = boxed_provider.downcast::<NullProvider>()
```

but I couldn't figure out how to do that. Specifically, I don't think we'll have any other providers that return true from `is_null` except `NullProvider`.

Providers that might use be hidden are things like `DebugProvider`. We don't allow it in production now, but it may be useful in the future to have available. If we do that, of course we wouldn't want to advertise it to users in the UI.

merino/src/docs/api.rs
mythmon: Oops. Probably better to keep it all `snake_case` for consistency.

merino-web/src/endpoints/suggest.rs
```diff
@@ -0,0 +1,153 @@
+//! Web handlers for the suggestions API.
```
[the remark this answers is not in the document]
mythmon: I did a fairly large amount of reshuffling, so I'm not surprised git didn't recognize the move. A lot of this file ended up going to `merino-web/providers.rs`

merino/src/docs/api.rs
pjenvey: 👍 The downcast impls I've seen tend to need unsafe and other trickery, so probably better to have a simpler `is_null` instead.

pjenvey (approved):
Just needs a small clippy fix per ci checks

mythmon:
Oh no. This test passes fine locally.

PR #141, opened 2021-09-22
doc: Convert general documentation to mdBook

mythmon:
I am of the opinion that while it is nice to have our docs generated straight from the code, it is too limiting in terms of the output it can make and how navigable we can make it for our documentation consumers (devs, leadership, ops). It was a nice experiment, but I don't think it worked out.

In the future I'd like to be able to extract documentation comments from Rust items to include in the mdBook, Sphinx-style, but for now I don't think that's possible.

Notably, the page detailing Merino's settings is mostly new, based loosely of doc comments that would have been strewn over many pages in the Rust Doc. This will certainly be harder to maintain, but I think the end result is worth it.

I'm making this as a draft to gather feedback. There are a couple PRs in flight right now that modify either the documentation or a major documented feature I'd like to wait until those land, and then integrate the changes into this PR before undrafting it.

Here's a screenshot of how the API documentation page looks like.

image

Compare that to the same documentation in the existing RustDoc version:

image


mythmon:
This needs https://github.com/mozilla/ci-docker-bases/pull/272

docs/ops.md
```diff
@@ -0,0 +1,215 @@
...
+  increasing verbosity), with an optional component that specifies the source of
+  the logs. For example `INFO,merino_web=DEBUG,reqwest=WARN` would set the
+  default log level to INFO, but would lower the level to `DEBUG` for the
+  `merino-web` crate and raise it to `WARN` for the reqwest crate.
+
+### Metrics
+
+Settings for Statsd/Datadog style metrics reporting.
+
+- `metrics.sink_address` (`MERINO_METRICS__SINK_ADDRESS`) - The IP and port to
```
pjenvey, line 85: Needs updating to sink_{host,port} per #142

docs/ops.md
```diff
@@ -0,0 +1,215 @@
...
+
+- `redis.url` (`MERINO_REDIS__URL`) - The URL to connect Redis at. Example:
+  `redis://127.0.0.1/0`.
+
+### Remote_settings
+
+Connection to Remote Settings. This is used by the Remote Settings suggestion
+provider below.
+
+- `remote_settings.storage_path` (`MERINO_REMOTE_SETTINGS__STORAGE_PATH`) - The
```
pjenvey, line 122: Killed by #133 in case you want to remove this line now

pjenvey (changes requested):
Just one or two things

docs/ops.md
mythmon: I'll handle that in #133.

pjenvey (approved):

PR #142, opened 2021-09-23
feat: make metrics sink address support hostnames

### LINKED ISSUE #137: Make metrics sink address support hostnames
It currently only supports IP address and ports, which is counter to what ops needs, and also counter to what our docs say.

Contile already supports this type of configuration, so we can use that as a template if needed.

**Acceptance Criteria**

* The metrics sink can be specified with a hostname and port like `dev-statsd-telegraf.influx.svc.cluster.local:8125`
* Metrics sink can still be specified using IP, like currently.
* There should be a test for the above

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1263)

Dexterp37:
This fixes [this JIRA ticket](https://mozilla-hub.atlassian.net/browse/CONSVC-1263).

Fixes #137

### DIFF
--- a/config/test.yaml
+++ b/config/test.yaml
@@ -8,7 +8,8 @@ redis:
   url: redis://localhost:6379
 
 metrics:
-  sink_address: "0.0.0.0:0"
+  sink_host: "0.0.0.0"
+  sink_port: 8125
 
 sentry:
   mode: debug
--- a/merino-integration-tests/src/utils/test_tools.rs
+++ b/merino-integration-tests/src/utils/test_tools.rs
@@ -91,8 +91,11 @@ where
 
     // Setup metrics
     assert_eq!(
-        settings.metrics.sink_address,
-        "0.0.0.0:0".parse().unwrap(),
+        settings.metrics.sink_host, "0.0.0.0",
+        "Tests cannot change the metrics sink host, since it is ignored"
+    );
+    assert_eq!(
+        settings.metrics.sink_port, 8125,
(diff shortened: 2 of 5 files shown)


config/test.yaml
```diff
@@ -8,7 +8,8 @@ redis:
   url: redis://localhost:6379
 
 metrics:
-  sink_address: "0.0.0.0:0"
+  sink_host: "0.0.0.0"
+  sink_port: 0
```
mythmon, line 12: I think that this should default to `8125`. `0.0.0.0` is a sort-of-sensical port for this to be on, but port 0 definitely doesn't make sense here.
```suggestion
  sink_port: 8125
```

mythmon (changes requested):

mythmon (approved):

merino-integration-tests/src/utils/test_tools.rs
```diff
@@ -91,8 +91,11 @@ where
 
     // Setup metrics
     assert_eq!(
-        settings.metrics.sink_address,
-        "0.0.0.0:0".parse().unwrap(),
+        settings.metrics.sink_host, "0.0.0.0",
+        "Tests cannot change the metrics sink host, since it is ignored"
+    );
+    assert_eq!(
+        settings.metrics.sink_port, 0,
```
Dexterp37, line 98:
```suggestion
        settings.metrics.sink_port, 8125,
```

PR #149, opened 2021-10-04
feat(suggest): Add timeout provider

### LINKED ISSUE #55: Add timeouts to suggestion providers
Suggestion providers should have a maximum time that they will wait for results. Requests that take longer than this configurable issue should time out and return no results.

**Acceptance Criteria**

* There is a suggestion provider that wraps another provider and returns an empty response if the request takes longer than a configurable amount of time.
* If a request times out, that should be visible in the API response.
* Tests should verify this
* A metric should be emitted when requests time out

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1143)
┆Epic: merino-milestone-1

mythmon:
- test(cache): Make memory locking test failures easier to read
- feat(suggest): Add timeout provider

Fixes #55

### DIFF
--- a/merino-suggest/src/timeout.rs
+++ b/merino-suggest/src/timeout.rs
@@ -1,0 +1,107 @@
+//! Tools to make sure providers don't  cache_status: todo!(), cache_ttl: todo!(), suggestions: todo!() take excessive amounts of time.
+
+use crate::{CacheStatus, SuggestionProvider, SuggestionResponse};
+use async_trait::async_trait;
+use merino_settings::providers::TimeoutConfig;
+use std::time::Duration;
+
+/// A combinator provider that returns an empty set of suggestions if the wrapped provider takes too long.
+pub struct TimeoutProvider {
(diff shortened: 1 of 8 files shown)


merino-suggest/src/timeout.rs
```diff
@@ -0,0 +1,107 @@
...
+impl SuggestionProvider for TimeoutProvider {
+    fn name(&self) -> String {
+        format!("timeout({})", self.inner.name())
+    }
+
+    async fn suggest(
+        &self,
+        query: crate::SuggestionRequest,
+    ) -> Result<crate::SuggestionResponse, crate::SuggestError> {
+        let inner_fut = self.inner.suggest(query);
```
tiftran, line 38: fut being future?

tiftran (approved):
looks good to me?

merino-suggest/src/timeout.rs
mythmon: Right.

PR #158, opened 2021-10-11
doc: Fix the link to the service docs

Dexterp37:
Documentation does not render correctly with the previous link.


mythmon:
I think that I fixed this in a different way in #157. Can you confirm that, @Dexterp37?

PR #162, opened 2021-10-13
chore: Add CODEOWNERS file

mythmon:
This should allow us to automatically request review from the appropriate groups.


pjenvey (approved):

mythmon:
Lets hope it works like we want it to!

PR #164, opened 2021-10-14
feat: implement a provider that filters results by keywords

### LINKED ISSUE #62: Add filter that prevents us from providing suggestions for configurable keywords
We want to make sure that some topics don’t get recommended by Merino. One technique we’ll use for this is to have a simple keyword filter that prevents sending suggestions that have certain keywords. We’ll want something that we can change without a redeploy _eventually_ but for now this can be configured with {{merino-settings}} and YAML files. The list of blocked words should not be commited to the repository, since we don’t want to publish a list of naughty words.

**Acceptance Criteria**

* Add a combinator-provider that filters suggestions from a subprovider
* Blocked keywords should be specified in settings
* Keywords should consist of an ID for the block pattern and a regex to match
* A metrics key should be incremented everytime a block pattern matches (for each suggestion), that specifies the ID of the pattern blocked.
* There should be tests for the above

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1136)
┆Epic: merino-milestone-1

Dexterp37:
This implements a new provider that, given the provided blocklist from the configuration files, filters the suggestions from the wrapped provider.

A counter telemetry is further recorded and sent whenever a filter matches a suggestion.

Fixes #62 

This got a bit bigger than expected, but most of the changes to the non provider files are due to propagating the `StatsdClient`.
Also note that the matching logic is very naive: can be further optimized if needed (e.g. using `RegexSet`s).


merino-settings/src/lib.rs
```diff
@@ -84,6 +84,9 @@ pub struct Settings {
 
     /// Settings to use when determining the location associated with requests.
     pub location: LocationSettings,
+
+    /// A blocklist to filter suggestions coming providers. Optional.
+    pub suggestion_blocklist: Option<HashMap<String, String>>,
```
mythmon, line 89: I don't think we should make this optional. This provider doesn't make sense without a blocklist.

merino-suggest/src/keyword_filter.rs
```diff
@@ -0,0 +1,223 @@
...
+    pub fn new_boxed(
+        blocklist: HashMap<String, String>,
+        inner: Box<dyn SuggestionProvider>,
+    ) -> Result<Box<Self>, SetupError> {
+        // Compile the provided blocklist regexes just once.
+        let mut compiled_blocklist: HashMap<String, Regex> = HashMap::new();
+
+        for filter in &blocklist {
+            if let Ok(r) = Regex::new(filter.1) {
+                compiled_blocklist.insert(filter.0.to_string(), r);
```
mythmon, line 32: How would you feel about naming the parts of the filter, instead of using `filter.0` and `filter.1`?

```suggestion
        for (filter_id, filter_regex) in &blocklist {
            if let Ok(r) = Regex::new(filter_regex) {
                compiled_blocklist.insert(filter_id.to_string(), r);
```

merino-suggest/src/keyword_filter.rs
```diff
@@ -0,0 +1,223 @@
...
+    /// Construct a new, boxed filter provider.
+    pub fn new_boxed(
+        blocklist: HashMap<String, String>,
+        inner: Box<dyn SuggestionProvider>,
+    ) -> Result<Box<Self>, SetupError> {
+        // Compile the provided blocklist regexes just once.
+        let mut compiled_blocklist: HashMap<String, Regex> = HashMap::new();
+
+        for filter in &blocklist {
+            if let Ok(r) = Regex::new(filter.1) {
```
mythmon, line 31: I think making filters case sensitive is pretty jarring to those that would be configuring the system. I think making the regex case insensitive would be good:

```suggestion
            if let Ok(r) = RegexBuilder::new(filter.1).case_insensitive(true).build() {
```

merino-web/src/providers.rs
```diff
@@ -84,6 +84,12 @@ async fn make_provider_tree(
         SuggestionProviderConfig::Fixed(fixed_config) => {
             FixedProvider::new_boxed(settings, fixed_config)?
         }
+
+        SuggestionProviderConfig::KeywordFilter(filter_config) => {
+            let inner = make_provider_tree(settings, filter_config.inner.as_ref()).await?;
+            KeywordFilterProvider::new_boxed(filter_config.suggestion_blocklist.clone(), inner)?
```
mythmon, line 90: We could give `KeywordFilterProvider` a reference to a metrics client here, maybe.

mythmon (dismissed):
Two thoughts:

1. Would using [`RegexSet`](https://docs.rs/regex/1.5.4/regex/struct.RegexSet.html) make sense, as a more efficient way to check many patterns at once? I expect we'll have a lot of these over time.
2. Instead of threading a metrics client through the suggest method's parameters, could it be given to providers at their creation time? Then only providers that want to engage in metrics would need the parameter.

merino-settings/src/lib.rs
Dexterp37: Whoops, this is actually bogus code: the blocklist is part of the provider configuration, not the general configuration.

merino-suggest/src/keyword_filter.rs
```diff
@@ -0,0 +1,243 @@
...
+    ) -> Result<crate::SuggestionResponse, crate::SuggestError> {
+        let mut results = self
+            .inner
+            .suggest(query)
+            .await
+            .unwrap_or_else(|_| SuggestionResponse::new(vec![]));
+
+        let mut reported_hits: HashMap<String, i64> = HashMap::new();
+        results.suggestions.retain(|r| {
+            let matches: Vec<_> = self.blocklist_rules.matches(&r.title).into_iter().collect();
```
ncloudioj, line 70: Do we really want to check the suggestion against _all_ the blocking rules here? Doing so could incur a performance impact when the ruleset is big. If we only care about blocking a suggestion if it matches _any_ rule, then it will be much faster to do `is_match` other than `matches` with the downside of not knowing the hit rule.

If we do want to test all of them and record which rules are hit, we might want to be mindful here as Regex matching could get CPU bound on a large ruleset, which could slow down the event loop and introduce an extra latency to the system.

ncloudioj (approved):
LGTM, r+ with a comment for your consideration.

Really liked this feature. I just realized that with a little extra handling, perhaps this could serve as a general-purpose suggestion filter, i.e. not only can it exclude entries from the result set with the blocklist, it could also include entries with the allowlist, although I am not sure how useful the latter could be in practice :)

pjenvey:
> I'm honestly not sure why the audit is failing now. Maybe because I touched the Cargo.lock ? The reported dependency was already being used.. @pjenvey or @jrconlin , any clue?

An older advisory was just applied retroactively to the older time crate: https://twitter.com/RustSec/status/1450115042075037702

I'm not sure how to deal with this one yet..

PR #187, opened 2021-10-27
[POC] feat: use actix-web parsing for accept language

### LINKED ISSUE #90: Use Actix Web to parse Accept-Language header
As part of CONSVC-107, we wrote code to manually parse the `Accept-Language` headers of incoming requests. Instead, let's use [Actix Web](https://docs.rs/actix-web/3.3.2/actix_web/http/header/struct.AcceptLanguage.html) to parse the header. An example of this on syncstorage can be found [here](https://github.com/mozilla-services/syncstorage-rs/blob/master/src/web/extractors.rs#L116).

Benefits to this approach:

* Reusing code in a widely-used library is less error-prone than maintaining the parsing code ourselves
* If the format for the `Accept-Language` header is ever changed, we would (hopefully) be able to rely on upstream libraries to implemented those changes (as opposed to having to keep up with such changes ourselves)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1183)

Dexterp37:
This is a proof-of-concept, **not meant to be merged.**

Fixes #90

### DIFF
--- a/merino-web/src/extractors.rs
+++ b/merino-web/src/extractors.rs
@@ -5,7 +5,15 @@ use std::str::FromStr;
 use crate::errors::{HandlerError, HandlerErrorKind};
 use actix_web::{
     dev::Payload,
-    http::{header, HeaderValue},
+    http::{
+        header::{
+            self,
+            Header,
+            Quality,
+            QualityItem
+        },
+        HeaderValue
+    },
     web::Query,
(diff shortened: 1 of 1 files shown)


mythmon:
@Dexterp37 thanks for making this proof of concept. I think for now I'm convinced we shouldn't switch over, at least until we hear back about the `*` parsing issue.

PR #193, opened 2021-11-01
feat: initialize suggestion providers at start up

### LINKED ISSUE #127: Set up the remote settings suggestion provider at server start
Currently the Merino Remote Settings suggestion provider waits until the first request per worker before syncing data from Remote Settings. While convenient, this also means that the first N requests are delayed while each worker independently loads the content. We should pre-load this data and share it between workers, if possible.

**Acceptance criteria**

* Load remote-settings data immediately after startup when the rs_adm provider is used
* Share the loaded data between Actix worker, or demonstrate that it is too hard to do reasonably

[https://github.com/mozilla-services/merino/issues/127](https://github.com/mozilla-services/merino/issues/127|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1247)
┆Epic: merino-milestone-1

jrconlin:
Moves initialization of suggestion providers to startup and shares the
initialized data between threads.

Closes #127


jrconlin:
TL:DR; moved the provider initialization to the async startup, outside of the app. Two tests are failing because the app isn't reading the remote settings URL because the data is already cached(?).

mythmon:
> Two tests are failing because the app isn't reading the remote settings URL because the data is already cached(?).

I think this is spot on. We'll need a way to resync the data from the server. Maybe we could add a `resync` method to `SuggestionProvider` that the tests could call? It may be nice to expose that in a `/debug` endpoint later too, so it isn't entirely wasted on tests.

jrconlin:
Yeah, mostly working out how the TestingTools are actually failing, but that might be a rabbit hole.

merino-integration-tests/src/suggest.rs
```diff
@@ -1,6 +1,9 @@
 //! Tests Merino's ability to make basic suggestions.
-#![cfg(test)]
+// Note, setting the following line will cause a compile error. This may be due to the
+// fact that the proc_macro may not be able to locate items.
+//#![cfg(test)]
 
+#![allow(unused_imports)]
```
mythmon, line 6: What import is unused, and if so why do we need it?

merino-settings/src/lib.rs
```diff
@@ -90,6 +90,12 @@ pub struct Settings {
     /// off, the suggest request object should be logged, but the
     /// search query should be blank.
     pub log_full_request: bool,
+
+    /// Only used for integration tests.
+    // note `#[cfg(test)]` will cause a compile error as the
+    // `merino_test` proc_macro will not be able to find this
+    // field.
+    pub test_changes: Option<Vec<String>>,
```
mythmon, line 98: I'd think this should be nested under the Remote Settings configuration object, instead of at the top level, or at least mention that it is for Remote Settings. Also some documentation about what it is and how to do it would be warranted.

merino-suggest/src/domain.rs
```diff
@@ -39,6 +39,7 @@ impl Proportion {
     }
 }
 
+/// Define helper impls for float values
```
mythmon, line 42: heads up: I think this might conflict with some other PR, but I forget which one.

merino-integration-tests/src/suggest.rs
```diff
@@ -1,6 +1,9 @@
 //! Tests Merino's ability to make basic suggestions.
-#![cfg(test)]
+// Note, setting the following line will cause a compile error. This may be due to the
+// fact that the proc_macro may not be able to locate items.
+//#![cfg(test)]
```
mythmon, line 4: Instead of removing this, we should move `setup_remote_settings_collection` into the testing tools file, since it is only used there now.

mythmon (changes requested):

merino-integration-tests/src/suggest.rs
jrconlin: There are a number, including:
```
crate::utils::test_tools::TestReqwestClient,
crate::TestingTools
anyhow::Result
merino_settings::providers::*
request::StatusCode
std::collections::*
```

As noted in the comment above, setting `#![cfg(test)]` causes the compile to fail because `merino-integration-tests::utils::test_tools::merino_test` cannot find `crate::suggest`, presumably because `crate::suggest` is not compiled as part of the non test build. So, instead, I enabled suggest, with the assumption that the remaining, non-test function `setup_remote_settings_collection(...)` is fine being part of the release build.

merino-integration-tests/src/suggest.rs
jrconlin: Ah, far better suggestion. thanks!

merino-integration-tests/src/suggest.rs
```diff
@@ -1,9 +1,10 @@
 //! Tests Merino's ability to make basic suggestions.
+// Note, setting the following line will cause a compile error. This may be due to the
+// fact that the proc_macro may not be able to locate items.
```
mythmon, line 3: I think this is just leftover that should be removed?

mythmon (approved):

merino-web/src/endpoints/suggest.rs
```diff
@@ -48,29 +48,18 @@ async fn suggest(
...
-                "suggester error"
-            );
-            HandlerError::internal()
-        })?;
-
     let extensions = request.extensions();
     let request_id: &Uuid = extensions
         .get::<RequestId>()
         .ok_or_else(HandlerError::internal)?;
+    let provider_ref = &provider.clone().0;
```
pjenvey, line 55: This isn't `provider_ref` it's `IdMulti` (the provider). Also the `Arc::clone` isn't needed here
```suggestion
    let provider = &provider.0;
```

merino-web/src/endpoints/providers.rs
```diff
@@ -22,21 +20,8 @@ pub fn configure(config: &mut web::ServiceConfig) {
...
-        .map_err(|error| {
-            tracing::error!(
-                ?error,
-                r#type = "web.suggest.setup-error",
-                "suggester error"
-            );
-            HandlerError::internal()
-        })?;
-
+    let provider = provider.0.clone();
```
pjenvey, line 24:
```suggestion
    let provider = &provider.0;
```

pjenvey (changes requested):

pjenvey (changes requested):
missed that other clone on the `SuggestionProviderRef` that could be removed

pjenvey (approved):

PR #194, opened 2021-11-01
Add per-provider cache keys

### LINKED ISSUE #140: Customize cache keys per provider
Currently, all suggestion requests use the same cache key, regardless of what we plan to do with that request. This made sense when we expected to have one primary provider, and expected that all providers would be roughly similar.

Now, we are planning to have fairly unique separate providers, and each provider would benefit from customized cache behavior (that is, exactly what [~accountid:60c36951c90cb2006878dcaf] proposed when this was first up for review).

We should provide a way to have "end providers" (those like `adm_rs`, `WikiFruit`, and a future Wikipedia provider) specify the cache keys that are relevant to their responses. This can then be passed up through the provider trees and used by caching providers. Combinator providers (like `Multi`) should pass that information up, and reformat it where neccessary.

**Acceptance Criteria**

* Cache keys are specified by end-providers
* Existing unit tests pass
* Combining multiple providers using combinators (like `Multi`) works as expected.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1267)
┆Epic: merino-milestone-1

mythmon:
- feat(suggest): Suggestion providers can their own cache keys
- perf: Specify minimum set of cache inputs needed for every provider

Fixes #140


merino-suggest/src/lib.rs
```diff
@@ -291,6 +291,31 @@ pub trait SuggestionProvider: Send + Sync {
...
         false
     }
+
+    /// Generate a set of cache inputs for a given query specific to this
+    /// provider. Any property of the query that affects how suggestions are
+    /// generated should be included.
+    ///
+    /// By default, all properties of the query are used, but providers should
+    /// narrow this to a smaller scope.
+    fn cache_inputs(&self, req: &SuggestionRequest, hasher: &mut blake3::Hasher) {
```
mythmon, line 301: I would have liked to keep `blake3` out of the "public" interface, but I couldn't find a way that I was happy with. Initially I tried to make this argument a closure that would take `&[u8]` that the provider could call multiple times. I couldn't get that to be both object safe for the trait and also lifetime safe.

I want to avoid `cache_inputs` just returning a `Vec` of strings or `&[u8]`s, because I think allocating multiple times for every provider for every request would get us in trouble.

mythmon: I also tried such silly things as `Arc<Box<dyn Fn(&[u8])>>` and `std::sync::mpsc::Channel`.

Dexterp37: No real clue on how to work around this either.

merino-suggest/src/lib.rs
```diff
@@ -291,6 +291,31 @@ pub trait SuggestionProvider: Send + Sync {
...
+        hasher.update(req.query.as_bytes());
```
Dexterp37, line 302: The blake3 docs mention a performance concern if the input buffer is less than 16KiB. Not sure we can do anything about it right now, but I was wondering if you had checked it.

mythmon: I assume you're talking about the docs for `Hasher::update`, here? https://docs.rs/blake3/1.1.0/blake3/struct.Hasher.html#method.update

I read that more as 8KiB and 16KiB buffers get faster since it can use wider SIMD instructions. I wouldn't take that to mean there is a performance _problem_ for shorter inputs. I don't think it would be faster to artificially pad this out to 16KiB just to get into this fast case (though I'd be happy to see benchmarks contrary). I think the docs there are warning against naively using `std::io::copy` for hashing large inputs, since `std::io::copy` has a limited buffer size.

When we chose Blake3 for this hash [in this ADR](https://mozilla-services.github.io/merino/adrs/adr_0002_hashing.html#benchmarking-results-for-hashing-128-byte-values), I benchmarked all the candidates using 128 byte values, and even in that case Blake3 was the fastest network-safe hash that we used.

Dexterp37: Got it, thanks.

Dexterp37 (approved):

merino-suggest/src/lib.rs
mythmon: @pjenvey do you have any clever ideas here? We can definitely ship this as is, but my JS-brain is unhappy about this, and I hoped there could be something better. :joy:

ncloudioj: Yeah, it's not ideal to expose this particular hasher type here. I wonder if we can make this a generic function, and then wrap the black::Hasher struct with a new type and implement the `write` method by simply making a `update` call. 

```rs
use std::hash::Hasher;

fn cache_inputs<T: Hasher>(&self, req: &SuggestionRequest, hasher: &mut T) {
   ...
}

struct Blake3Hasher(blake::Hasher);

impl Hasher for Black3Hasher {
  fn write(&mut self, bytes: &[u8]) {
    self.update(bytes);
  }
}
```

A little bit boilerplate, but it works with the standard Hasher trait, which allows us to change it in the future if needed.

mythmon:
```rust
fn cache_inputs<T: Hasher>
```

We can't make `cache_inputs` have any generic parameters, since `SuggestionProvider` is used primarily as a trait object, and generics aren't object safe. 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a5070587d7766b5bfe77dd6e440f795a
https://doc.rust-lang.org/book/ch17-02-trait-objects.html#trait-objects-perform-dynamic-dispatch

If we could do generics, this could use a closure directly with

```rust
fn cache_inputs<F: Fn(&[u8])>(&self, req: &SuggestionRequest, closure: F);
```

ncloudioj: Ah, now I know what this object safety is all about.

And we cannot do `hasher: impl Hasher` in the trait implementation; `Box<dyn Hasher>` is not cheat either. 🤷‍♂️

mythmon: `Box<dyn Hasher>` isn't too bad. I tried something similar to that with closures. I wasn't able to get the lifetimes sorted with the closure, but maybe I could with `Box<dyn Hasher>`

pjenvey: Could also use just `hasher: &mut dyn Hasher` here for the same dynamic dispatch without the Box.

`blake3::Hasher` even implements `std::io::Write` which could be used for the trait here but its `write` method is a little annoying, returning an io Result.

mythmon: `Box<dyn Hasher>`, compared to `Box<Dyn Fn(&[u8])>`, has the advantage that I can add a `finish` method to it. That fixes the major problem I had with the closure, because I couldn't easily get the blake3 hasher back once I gave it to the closure.

mythmon: `&mut dyn Hasher` works great. I had forgotten that would work. I forget sometimes that you can do unsized stuff in other ways that `Box`.

ncloudioj: Right, dynamic dispatching works for references too.

ncloudioj (approved):
🚢

merino-suggest/src/lib.rs
```diff
@@ -291,6 +291,31 @@ pub trait SuggestionProvider: Send + Sync {
...
+    /// Generate a set of cache inputs for a given query specific to this
+    /// provider. Any property of the query that affects how suggestions are
+    /// generated should be included.
+    ///
+    /// By default, all properties of the query are used, but providers should
+    /// narrow this to a smaller scope.
+    fn cache_inputs(&self, req: &SuggestionRequest, cache_inputs: &mut dyn CacheInputs) {
+        cache_inputs.add(req.query.as_bytes());
+        cache_inputs.add(&[req.accepts_english as u8]);
+        cache_inputs.add(format!("{:?}", req.country).as_bytes());
```
ncloudioj, line 304: We could replace those `format!`s with `cache_inputs.add(req.country.as_deref().unwrap_or("None").as_bytes())`.

I know it's messy, but my "let's-save-one-allocation" OCD just kicked in. 😂

PR #204, opened 2021-11-04
Download suggestions from Kinto in load tests

### LINKED ISSUE #199: Pull suggest queries from RS into load tests
The load tests expect an `InstantSuggest_Queries_*.json` file from the source-data directory in the `quicksuggest-rs` for generating suggest queries in load testing tasks.

Pull the same RS collection with the source data as the target Merino deployment, ideally only once for the load tests and then sync between workers and use that instead of a local JSON file on disk.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1335)
┆Epic: merino-milestone-1

hackebrot:
This is work in progress. 🚧 

I will write more documentation for the updated load test suite before we merge this pull request. I am able to run the load tests locally using the Kinto production endpoint and the default bucket and collection in Merino's settings. The suggestions data is only downloaded once on the Locust `MasterRunner` when the load test run is started and a random selection of suggestions is then stored on the workers to be later used in the tasks when sending requests.

@mythmon @ncloudioj can you please take a look at the Kinto integration? I tried to mimic Merino's handling of records and attachments and the kinto-http Python SDK doesn't seem to support attachments out of the box. I'm sure there are a few things that could be more robust, but I'm focused on getting us to a place where we can start running load tests asap.

Resolve #199


test-engineering/load-tests/tests/kinto.py
```diff
@@ -0,0 +1,60 @@
...
+    suggestions = {}
+
+    for record in data_records:
+        attachment_url = f"{attachments_base_url}{record['attachment']['location']}"
+
+        response = requests_session.get(attachment_url)
+
+        if response.status_code != 200:
+            # Ignore unsuccessful requests for now
+            continue
```
mythmon, line 46: It's probably worth logging a warning here. If this is happening we should probably see if it affects Firefox.

test-engineering/load-tests/tests/locustfile.py
```diff
@@ -24,6 +30,61 @@
...
+# The number of random suggestions stored on each worker
+RS_SUGGESTIONS_COUNT: int = 100
+
+# This will be populated on each worker and
+RS_SUGGESTIONS: List[Dict] = []
+
+
+@events.test_start.add_listener
+def on_locust_test_start(environment, **kwargs):
+    """Download suggestions from Kinto and store random suggestions on workers."""
```
mythmon, line 51: Why do workers receive a random set, instead of all of them suggestions?

test-engineering/load-tests/tests/kinto.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # Each attachment is a list of suggestion objects
+        # Each suggestion objects contains a list of keywords
+        attachment = json.loads(response.text)
```
mythmon, line 50: Requests has a method specifically for getting the JSON body of the response: https://docs.python-requests.org/en/latest/user/quickstart/#json-response-content

```suggestion
        attachment = response.json()
```

mythmon (commented):
The Kinto interactions look correct to me.

test-engineering/load-tests/tests/kinto.py
hackebrot: Good catch! Thanks! 👍🏻

test-engineering/load-tests/tests/locustfile.py
hackebrot: I did this mainly for debugging purposes during development. I'll update this to store all suggestions.

test-engineering/load-tests/tests/kinto.py
hackebrot: I've added a log entry on level `ERROR` for this.

ashrivastava-qa (approved):
LGTM! 👍

hackebrot:
I just reworded one of the commit messages to not use the `test:` prefix.

hackebrot:
Thanks for the reviews @mythmon and @ashrivastava-qa! 🚀

PR #205, opened 2021-11-05
feat: Add duration-us metrics to RedisCache, MemoryCache, and RemoteSettings providers

mythmon:


merino-adm/src/remote_settings/mod.rs
```diff
@@ -213,16 +221,24 @@ impl SuggestionProvider for RemoteSettingsSuggester {
...
                 _ => vec![],
             }
         } else {
             vec![]
-        };
+        });
+
+        self.metrics_client
+            .histogram(
+                "adm.provider.duration-us",
```
Dexterp37, line 236: Would you kindly document this histogram [here](https://github.com/mozilla-services/merino/blob/main/docs/data.md)?

merino-cache/src/memory.rs
```diff
@@ -207,33 +218,51 @@ impl SuggestionProvider for Suggester {
...
+                        suggestions: Vec::new(),
+                    });
+                } else {
+                    // handle cache miss or stale cache
+                    let lock = LOCK_TABLE.add_lock(&key, self.default_lock_timeout);
+                    let mut response = self
+                        .inner
+                        .suggest(query)
+                        .await?
+                        // Todo, cache status should be a vec.
```
Dexterp37, line 236: Should this be turned into an issue on GH/Jira Ticket and have it referenced here?

merino-cache/src/memory.rs
```diff
@@ -207,33 +218,51 @@ impl SuggestionProvider for Suggester {
...
-            if LOCK_TABLE.is_locked(&key) {
-                // there's a fetch already in progress. Return empty for now.
-                return Ok(SuggestionResponse {
-                    cache_status: CacheStatus::Hit,
-                    cache_ttl: None,
-                    suggestions: Vec::new(),
-                });
+            if rv.is_none() {
+                if LOCK_TABLE.is_locked(&key) {
+                    // there's a fetch already in progress. Return empty for now.
```
Dexterp37, line 223:
```suggestion
                    // There's a fetch already in progress. Return empty for now.
```

merino-cache/src/memory.rs
```diff
@@ -207,33 +218,51 @@ impl SuggestionProvider for Suggester {
...
+                    rv = Some(SuggestionResponse {
+                        cache_status: CacheStatus::Hit,
+                        cache_ttl: None,
+                        suggestions: Vec::new(),
+                    });
+                } else {
+                    // handle cache miss or stale cache
```
Dexterp37, line 230:
```suggestion
                    // Handle cache miss or stale cache.
```

merino-cache/src/memory.rs
```diff
@@ -207,33 +218,51 @@ impl SuggestionProvider for Suggester {
...
-                let expiration = now + *cache_ttl;
-                tracing::debug!(?now, ?expiration, "inserting into cache");
-                self.items
-                    .insert(key.clone(), expiration, response.suggestions.clone());
-            });
-            Ok(response)
+            if let Some(response) = rv {
+                self.metrics_client
+                    .histogram_with_tags(
+                        "cache.memory.duration-us",
```
Dexterp37, line 255: Would you kindly document this as well?

merino-cache/src/redis/mod.rs
```diff
@@ -363,44 +372,48 @@ impl SuggestionProvider for Suggester {
...
-                    response.with_cache_status(CacheStatus::Miss)
-                } else {
-                    debug_assert!(matches!(cache_result, CacheCheckResult::ErrorAsMiss));
-                    tracing::debug!(%key, "cache error treated as miss");
-                    response.with_cache_status(CacheStatus::Error)
-                }
+            suggestions
+        } else if rlock.is_locked(&key).await? {
+            tracing::debug!(%key, "cache updating...");
+            // A "pending" review may not yet have content (e.g. it's the initial lookup), otherwise it's a "Hit"
```
Dexterp37, line 386:
```suggestion
            // A "pending" review may not yet have content (e.g. it's the initial lookup), otherwise it's a "Hit".
```

merino-cache/src/redis/mod.rs
```diff
@@ -363,44 +372,48 @@ impl SuggestionProvider for Suggester {
...
+                tracing::debug!(%key, "cache error treated as miss");
+                response.with_cache_status(CacheStatus::Error)
+            }
+        } else {
+            SuggestionResponse::new(Vec::new()).with_cache_status(CacheStatus::Error)
+        };
+
+        self.metrics_client
+            .histogram_with_tags(
+                "cache.redis.duration-us",
```
Dexterp37, line 411: Would you kindly document this as well?

Dexterp37 (changes requested):

Dexterp37:
@mythmon since you're touching caching here, would you mind also recording the metrics used for computing the following metrics:

- Suggestions cache hit ratio (this requires both cache hit and cache miss I guess)
- Suggestions cache size

These were requested by @hackebrot to be on the load testing dashboard [see the doc](https://docs.google.com/document/d/1QxEcQJGPf-BPzY82ZHnTr6b-U4qVlxQA3LmZH-C3INw/edit#).

I'm also happy to do this as a follow-up of this PR if you don't have time.

mythmon:
I don't think we have the needed metrics to monitor cache size or cache hit ratio right now, though I agree we should. I think we should push that off to another PR.

merino-cache/src/memory.rs
mythmon: I'm going to just take the comment out for now; I think we should see how we end up using the caches before we invest more into this. The possibilities I can think of already are, which are not mutually exclusive:

1. We only ever use one cache per provider-chain, so we only need a single cache status
2. We often pair memory and redis (or another) cache, and want to report cache status for both
3. We often use Multi or IdMulti, which may want to report cache statuses from multiple providers independently
4. We want to report cache status in metrics or logs
5. We want to also include the full cache picture in the HTTP responses

Once we know more about what we want/need out of this area, lets revisit this.

merino-adm/src/remote_settings/mod.rs
```diff
@@ -213,16 +221,24 @@ impl SuggestionProvider for RemoteSettingsSuggester {
...
             match self.suggestions.get(&request.query) {
                 Some((_, suggestion)) => vec![suggestion],
                 _ => vec![],
             }
         } else {
             vec![]
-        };
+        });
+
+        self.metrics_client
```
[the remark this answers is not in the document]
mythmon: Ah, good catch. I think tagging it is the right call. I agree that this provider is going to be very quick anyways though.

pjenvey (approved):

Dexterp37 (approved):
Thanks!

PR #219, opened 2021-11-17
feat: Add sub-microsecond precision using floats for timer metrics

mythmon:


merino-suggest/src/metrics.rs
```diff
@@ -0,0 +1,49 @@
...
+use cadence::{Histogram, Histogrammed, MetricBuilder, MetricResult};
+use std::time::Duration;
+
+/// Trait for recording timer values with additional precision.
+///
+/// Time data is stored as a histogram.  Statistical distribution is calculated
+/// by the server. Times will be stored a number of milliseconds, with
+/// fractional values used to represent nanoseconds.
+///
+/// Only `Duration types are valid.
```
pjenvey, line 12:
```suggestion
/// Only `Duration` types are valid.
```

pjenvey (changes requested):

mythmon:
The check failure is `cargo audit`. I've made #220 to handle that.

pjenvey (approved):

Dexterp37:
Do we need to update the data docs ?

mythmon:
The data docs currently say

> A histogram that records the amount of time, in microseconds, that the adM Remote Settings provider took to generate suggestions.

That doesn't specify what the precision of those microseconds are, so I don't think we need to update the docs.

mythmon:
Rebased to pick up CI changes.

Issue #218, opened 2021-11-17
Verify Remote Settings content signatures


data-sync-user:
When we switched from using the `remote-settings-client` Rust library to using Remote Settings directly (CONSVC-1258), we lost the signature verification of the downloaded collection.

In the November RRA (CONSVC-1337), Steph recommend either re-adding this signature verification before the Beta test, or having a timeline for when we would be re-adding the verification.

This signature verification is a defense-in-depth protection that prevents scenarios like an attacker redirecting Merino to pull from a malicious data source by changing its configuration.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1347)
┆epic: merino-beta-experiment

data-sync-user:
➤ Alessio Placitelli commented:

As decided in the Engineering meeting, I’m going to talk to leplatrem to see if we could make the changes required to use remote-settings-client in Merino instead of writing the signature verification code ourselves (since the latter would require security review).

data-sync-user:
➤ Alessio Placitelli commented:

I talked to Mathieu Leplatre and Michael Cooper today: we agreed that working on remote-settings-client and using it to verify the content signatures is the path of least resistance (compared to rolling out our own verification system), and it has the advantage of unblocking usage of the remote-settings-client library to other teams.

The rough plan we agreed on with mythmon:

  * Create a branch on Merino and re-integrate remote-settings-client in the current form
  * Implement attachments support in remote-settings-client
  * Highlight any friction point of the library usage and tackle them as needed

In order for the remote-settings-client library to be finally used, a security review must be performed on that library (Mathieu Leplatre to drive this)

data-sync-user:
➤ Alessio Placitelli commented:

This is a blocker for rolling out to the Release population, see this thread ( https://mozilla.slack.com/archives/G01NU8LUXPH/p1637601413385500?thread_ts=1637600516.381000&cid=G01NU8LUXPH ). As per the current plans, rollout to release will happen on January 11th.

data-sync-user:
➤ Michael Cooper commented:

An update to this. Although this defense in depth measure is useful, bringing down the entire infrastructure because of a signature bug is not. Our history with content signature shows us that having it automatically stop features hurts us more often than it helps us. I think this is doubly true in the server-to-server case of Merino.

[...]

PR #232, opened 2021-11-24
Integrate remote-settings-client

### LINKED ISSUE #221: Integrate remote-settings-client in Merino (in a branch)
**Acceptance criteria**

* Merino should compile 
* The remote-settings-client should be able to connect to the RS servers (even without downloading attachments)
* All the non RS related tests should pass.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1374)

Dexterp37:
This fixes #221 and contributes to #218 .

This does not currently enable signature verification, but exclusively replaces the custom rs implementation with the remote-settings-client one.


merino-adm/src/remote_settings/mod.rs
```diff
@@ -43,19 +42,28 @@ impl RemoteSettingsSuggester {
...
-                .as_ref()
-                .unwrap_or(&settings.remote_settings.default_bucket)
-                .clone(),
-            config
-                .collection
-                .as_ref()
-                .unwrap_or(&settings.remote_settings.default_collection)
-                .clone(),
-        )?;
+        let mut remote_settings_client = remote_settings_client::Client::builder()
```
pjenvey, line 45: Previously we were utilizing a `FileStorage`, e.g. `remote_settings_client.storage(Box::new(FileStorage..`, do we want to do that again? cc @mythmon

mythmon: Yes, if we don't we'll be re-downloading attachments at every sync.

merino-adm/src/remote_settings/reqwest_client.rs
```diff
@@ -0,0 +1,77 @@
...
+impl RsRequester for ReqwestClient {
+    async fn get(&self, url: RsUrl) -> Result<RsResponse, ()> {
+        match self
+            .reqwest_client
+            .get(url.clone())
+            .timeout(std::time::Duration::from_secs(3))
+            .send()
+            .await
+            .and_then(Response::error_for_status)
+            .context(format!("Fetching records from remote settings: {}", url))
```
mythmon, line 40: This is used for more than just fetching records. We could change this log line to "Performing HTTP request for Remote Settings", or similar.

merino-adm/src/remote_settings/reqwest_client.rs
```diff
@@ -0,0 +1,77 @@
...
+                let body = match response.bytes().await {
+                    Err(e) => {
+                        tracing::error!(
+                            "ReqwestClient - unable to parse response body. {:?}",
+                            e.to_string()
+                        );
+                        return Err(());
+                    }
+                    Ok(b) => b,
+                };
```
mythmon, line 67: Would something like 

```rust
let body = response.bytes().await
	.map_err(|err| {
		tracing::error!(...);
		()
	})?;
```

Work? Using `map_err` to both log the messages and transform the error into `()` helps re-enable the Try operator, while keeping the same behavior.

merino-integration-tests/src/utils/test_tools.rs
```diff
@@ -235,7 +235,13 @@ impl TestReqwestClient {
 async fn setup_remote_settings_bucket(server: &str) -> Result<(String, String)> {
     let reqwest_client = reqwest::Client::new();
     let bucket_info: serde_json::Value = reqwest_client
-        .post(format!("{}/v1/buckets/", server))
+        // Unfortunately `/dev/kinto.ini` does not support `/buckets/*`
+        // in `kinto.changes.resources`, so we should use the same bucket
+        // (i.e. `/buckets/main`) to create our collections in and make sure
+        // remote settings behave as expected. We use PUT here to overwrite
+        // the bucket transparently each time, since we don't really care
+        // about the stored metadata.
```
mythmon, line 243: Will calling `PUT` here delete the underlying collections? This method gets called as a part of parallel test runs, so we don't want to delete the collections used by other tests. The integration tests are one of the primary users of `/dev/kinto.ini`, so we can change that to support the old method, if that's possible.

mythmon (changes requested):

merino-integration-tests/src/utils/test_tools.rs
Dexterp37: Good point. I restored the POST

mythmon (approved):

merino-integration-tests/src/utils/test_tools.rs
leplatrem: > Will calling PUT here delete the underlying collections?

No it won't

PR #227, opened 2021-11-22
Reorganize test-engineering files

hackebrot:
This pull request moves the tests we formerly referred to as integration and later system tests to a new sub directory in the `test-engineering` directory named `contract-tests`, since that most accurately describes the kind of coverage they provide. 🤖 

The files are now properly separated from the load tests in `tests-engineering/load-tests`.

PR #235, opened 2021-11-24
feat(adm): Make suggestion score configurable

### LINKED ISSUE #234: Update the suggestion score in Merino
Merino now assigns a fixed score (0.2) to the suggestions of Remote Settings, which is exactly the same as what Firefox assigns to the RS suggestions. We should use a higher score in Merino so that Merino suggestions take precedence over RS.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1392)

mythmon:
Also changes the default from 0.2 to 0.3.

Fixes #234.


ncloudioj (approved):

mythmon:
Merged in main to pick up the change from `integration-tests` to `contract-tests` in #227.

Issue #224, opened 2021-11-18
Intermittent test failure with RS bucket creation


ncloudioj:
Seeing this kinda panics both locally and in CI:

```shell
thread 'suggest::suggest_providers_can_be_filtered' panicked at 'Error creating bucket: reqwest::Error { kind: Request, url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("localhost")), port: Some(8888), path: "/v1/buckets", query: None, fragment: None }, source: hyper::Error(IncompleteMessage) }', merino-integration-tests/src/utils/test_tools.rs:78:10
```

After a preliminary investigation, there appears to be a race condition in `test_tools::merino_test()` when running tests in multiple threads, the RS setup calls could interfere with each other.

PR #246, opened 2021-11-25
fix(api): Split advertisers from the providers

### LINKED ISSUE #245: The “advertiser” data from the telemetry pings have the “adm” value instead of the sponsored or non sponsored advertiser
**[Affected Versions]:**
- Firefox Nightly 96.0a1 (Build ID: 20211124155224)
- Firefox Beta 95.0b11 (Build ID: 20211123190150)
- Firefox Release 94.0.2 (Build ID: 20211117154346)

**[Affected Platforms]:**
- Windows 10 x64
- macOS 10.15.7
- Ubuntu 20.04 x64

**[Prerequisites]:**
- Have the "browser.urlbar.merino.enabled" set to true.
- Have the "browser.urlbar.merino.endpointURL" set to "https://stage.merino.nonprod.cloudops.mozgcp.net/api/v1/suggest".
- Have the "browser.urlbar.quicksuggest.remoteSettings.enabled" set to "false".
- Have the "browser.urlbar.quicksuggest.dataCollection.enabled" set to "true"
- Have the "browser.ping-centre.log" set to "true".
- Have the "browser.search.region" set to "US".
- Have the browser console opened.

**[Steps to reproduce]:**
1. Open the latest Nightly/Beta build with the profile from prerequisites.
2. Open a New Tab page and type "ebay".
3. Click the Sponsored result displayed on the last position.
4. Observe the impression and click pings registered in the browser console.

**[Expected results]:**
- Both pings contain the “advertiser” data with “ebay” value. 

**[Actual results]:**
- Both pings contain the “advertiser” data with “adm” value. 

**[Notes]:**
- For the Remote Settings results the " the “advertiser” data registers the correct values:
1. “ebay” for the sponsored advertiser
2. “wikipedia” for the non sponsored advertiser
- Not sure if this issue is related to #242.
- Attached is a screen recording of the issue:

advertiser

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1397)

ncloudioj:
This fixes #245.

We treat `provider` and `advertiser` as the same thing in the current API. Turns out that is insufficient to cover the following scenario: a `provider` could provide suggestions for multiple  `advertiser`s. For example: provider "adM" could provide sponsored suggestions for "advertiser-foo", "advertiser-bar".

As for this issue, the `provider` field was successfully populated when Merino loaded suggestions from Remote Setting. But later, it got overwritten by `IdMulti` at [here](https://github.com/mozilla-services/merino/blob/main/merino-suggest/src/providers/id_multi.rs#L92). As a result, all the suggestions now had the `adm` (i.e. the name of the `IdMulti` provider) as the `provider`, and that response was sent back to Firefox with the incorrect `advertiser`.


docs/api.md
```diff
@@ -111,11 +111,11 @@ The response will be a JSON object containing the following keys:
   [Interaction Pings](#interaction-pings), below. This field may be null, in
   which case no click ping is required for this suggestion provider.
 
-- `provider` - A string that identifies the source of this suggestion. This can
-  be used along with `block_id` to uniquely identify this suggestion. It is not
-  intended to be directly displayed to the user.
+- `provider` - A string that identifies the provider of this suggestion, such as
```
Dexterp37, line 114: If this is not meant to be displayed, why are we reporting it?

docs/api.md
```diff
@@ -111,11 +111,11 @@ The response will be a JSON object containing the following keys:
...
+  "adM". This field is not intended to be directly displayed to the user.
 
-- ~~`advertiser`~~ - A deprecated alias of `provider`.
+- `advertiser` - The name of the advertiser, such as "Nike". Note that a `provider`
+  could have multiple `advertiser`s. This field will be displayed to the user.
```
Dexterp37, line 118: `This field will be displayed to the user`: that's up to the consumer of the API, right? Is the consumer that can decide whether or not it _will_ be displayed. Should we drop this sentence at all? It makes more sens eto live in Firefox :)

merino-suggest/src/lib.rs
```diff
@@ -184,9 +184,12 @@ pub struct Suggestion {
     #[serde_as(as = "DisplayFromStr")]
     pub click_url: Uri,
 
-    /// The name of the advertiser associated with this suggestion.
+    /// The name of the provider associated with this suggestion.
     pub provider: String,
 
+    /// THe name of the advertiser associated with this suggestion.
```
Dexterp37, line 190:
```suggestion
    /// The name of the advertiser associated with this suggestion.
```

Dexterp37 (changes requested):
Looks good overall, with the requests addressed! Please let's take a look at the failing CI tests as well.

docs/api.md
ncloudioj: We're including it in the API response because Firefox could use it in the future, for instance, performing different suggestion handling on different providers. Though that is largely irrelevant to the Firefox users, so they won't see it.

docs/api.md
ncloudioj: Good point. Yeah, it's up to the consumers to determine whether or not to show it. I will drop it.

Dexterp37:
@hackebrot do you have any clue about why tests are failing here? Was anything related to kinto recently changed?

Dexterp37 (approved):
The code in this PR looks good to me. Let's not merge just yet, let's wait for #248 . If CI in there fails as well, then chances are testing is broken and failures are not related to this. Otherwise, we should definitely take another look before merging.

hackebrot:
Hi @Dexterp37! 👋🏻 

I reviewed the changes in this pull request:

* the `contract-tests` failure on 987afd5f9f22b3cc01fd7518b17946dcb6320c5d is expected as the formerly deprecated alias of `provider` named `advertiser` is now set to the new value `test_advertiser` instead of the same values as provider which is `test`
* the `build-and-test` failure on ea88ed28a84482a7677c762962492bca1b8d07fb is not expected. I see `Failed to set up the Remote Settings bucket` in the logs for the `cargo test` step.

Could that be related to #224, @ncloudioj? 🤖

ncloudioj:
> Could that be related to #224, @ncloudioj? 🤖

I don't think this one is related to #224, as that was an intermittent failure caused by the race condition across the test cases of the integration test. From the error message, looks like the Kinto service in the composer had this intermittent error, which prevented all the integration tests from running.

I will file a new issue for that with details.

PR #249, opened 2021-11-26
chore: tag 0.4.3

ncloudioj:
r? @jbuck

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+<a name="0.4.3"></a>
+## 0.4.3 (2021-11-26)
+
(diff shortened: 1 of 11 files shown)


jbuck:
looks like this is failing some tests, and I don't think I can actually review this patch since it'll require review from a code owner

ncloudioj:
> I don't think I can actually review this patch since it'll require review from a code owner

Right, nvm. The dev build succeeded, I think we can use that for QA for now. 

> looks like this is failing some tests

Yeah, that was a known intermittent failure (#250) with the dependency in Kinto.

pjenvey (approved):

PR #258, opened 2021-12-06
docs: Fix typos in release-template.md

ncloudioj:


.github/ISSUE_TEMPLATE/release-template.md
```diff
@@ -18,7 +18,7 @@ assignees: ""
   - If this is a new major/minor release,
     `git checkout -b release/{major}.{minor}` to create a new release branch.
   - If this is a new patch release:
-    - `git checkout release/{major}.{minor}`
+    - `git checkout release/{major}.{minor}.{patch}`
```
mythmon, line 21: This wasn't a typo. The intent was that dot releases would be bug fix releases we cherry picked back onto a release branch.

That being said, we aren't doing that, in practice, so I agree with this change, just not the description of the commit message.

mythmon (approved):

.github/ISSUE_TEMPLATE/release-template.md
ncloudioj: Sounds good.

Looks like Contile already updated its release template with the same changes. Also, it added the `.0` for the major/minor release, which I think makes sense.

ncloudioj: Merged the major/minor and the patch cases into one.

Issue #259, opened 2021-12-06
Intermittent test failure with ReqwestClient


ncloudioj:
This failure was intermittently observed both in CI and locally with the different test cases.

Case I:
```
ERRORmerino_adm::remote_settings::reqwest_client: ReqwestClient - unable to submit GET request. "Performing HTTP request for Remote Settings: http://localhost:8888/buckets/monitor/collections/changes/changeset?_expected=0", type: "adm.remote-settings.reqwest.get-failed"
    at merino-adm/src/remote_settings/reqwest_client.rs:45

thread 'suggest::suggest_adm_rs_works_empty' panicked at 'Could not create providers: There was a network error while setting up this suggestions provider
``` 

Case II:
```
ERRORmerino_adm::remote_settings::reqwest_client: ReqwestClient - unable to submit GET request. "Performing HTTP request for Remote Settings: http://localhost:8888/", type: "adm.remote-settings.reqwest.get-failed"
    at merino-adm/src/remote_settings/reqwest_client.rs:45
    in merino_adm::remote_settings::sync with suggestion_score: Proportion(1216271616)

thread 'suggest::suggest_adm_rs_score_is_configurable' panicked at 'Could not create providers: There was a network error while setting up this suggestions provider
```

pjenvey:
We're only seeing the outtermost `context` of the error in this log message, the inner isn't shown. anyhow docs suggest this might show the original cause but I haven't tested it yet:

https://github.com/mozilla-services/merino/compare/feat/259

pjenvey:
Per CI it indeed does, a hyper IncompleteMessage in this case:

```
  Dec 07 00:37:11.813 ERRORmerino_adm::remote_settings::reqwest_client: ReqwestClient - unable to submit GET request. Error {
    context: "Performing HTTP request for Remote Settings: http://localhost:8888/buckets/main/collections/g0nu4ZWS/changeset?_expected=1638837431756",
    source: reqwest::Error {
        kind: Request,
        url: Url {
            scheme: "http",
            cannot_be_a_base: false,
            username: "",
            password: None,
            host: Some(
                Domain(
                    "localhost",
                ),
            ),
            port: Some(
                8888,
            ),
            path: "/v1/buckets/main/collections/g0nu4ZWS/changeset",
            query: Some(
                "_expected=1638837431756",
            ),
            fragment: None,
        },
        source: hyper::Error(
            IncompleteMessage,
        ),
    },
[...]

PR #260, opened 2021-12-07
fix(adm): Include `/v1` on the end of the remote settings URL we give to the RS client

mythmon:
The default value that remote-settings-client uses for the server includes the /v1 suffix, but ours doesn't. Add it back in so that we don't confuse the client.

### DIFF
--- a/config/test.yaml
+++ b/config/test.yaml
@@ -9,7 +9,7 @@ redis:
 
 remote_settings:
   # This is the docker-compose provided server, in ./dev/.
-  server: "http://localhost:8888"
+  server: http://localhost:8888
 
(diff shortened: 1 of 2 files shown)


pjenvey (approved):

mythmon:
This is something I think might help #259. If it does, I'm in the "I have no idea how this ever worked" camp (and yet, it works on my machine, even without this change).

mythmon:
image

I'm going to force-push to remove those commits, but I think that this means it is working better now.

ncloudioj:
Ugh, this failure still happened to me with this patch. There is something else around it.

PR #261, opened 2021-12-07
feat: don't hide inner errors in the log message

pjenvey:
Issue #259


mythmon (approved):

ncloudioj:
Ha, looks like Kinto does the auto-redirection if the request URL doesn't include `v1`. The error log shows the `path` actually starts with `/v1`.

PR #262, opened 2021-12-07
test: Fix intermittent failures with ReqwestClient

### LINKED ISSUE #259: Intermittent test failure with ReqwestClient

ncloudioj:
This fixes #259.

The issue was caused by the connection pool in Reqwest/Hyper. See more details in [here](https://github.com/hyperium/hyper/issues/2136#issuecomment-589345238).

Since `ReqwestClient` is only used by RS suggester, and it is not making constant network calls to RS. Disabling the connection pool should not be a big deal, but I will let @mythmon @Dexterp37 decide if this is acceptable or shall we keep it as is and add the retry to the API.

### DIFF
--- a/merino-adm/src/remote_settings/reqwest_client.rs
+++ b/merino-adm/src/remote_settings/reqwest_client.rs
@@ -44,8 +44,8 @@ impl RsRequester for ReqwestClient {
             Err(e) => {
                 tracing::error!(
                     r#type = "adm.remote-settings.reqwest.get-failed",
-                    "ReqwestClient - unable to submit GET request. {:?}",
-                    e.to_string()
+                    "ReqwestClient - unable to submit GET request. {:#?}",
+                    e
                 );
                 Err(())
             }
@@ -61,8 +61,8 @@ impl RsRequester for ReqwestClient {
                 let body = response.bytes().await.map_err(|err| {
                     tracing::error!(
                         r#type = "adm.remote-settings.reqwest.parsing-failed",
-                        "ReqwestClient - unable to parse response body. {:?}",
-                        err.to_string()
+                        "ReqwestClient - unable to parse response body. {:#?}",
+                        err
                     );
                 })?;
 
--- a/merino-adm/src/remote_settings/reqwest_client.rs
+++ b/merino-adm/src/remote_settings/reqwest_client.rs
@@ -4,7 +4,7 @@
 
 //! An HTTP client implementation to use with the remote-settings-client.
 
-use anyhow::Context;
+use anyhow::{Context, Error};
 use async_trait::async_trait;
 use remote_settings_client::client::net::{
     Headers as RsHeaders, Requester as RsRequester, Response as RsResponse, Url as RsUrl,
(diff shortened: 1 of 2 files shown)


merino-adm/src/remote_settings/reqwest_client.rs
```diff
@@ -20,10 +20,13 @@ pub struct ReqwestClient {
 
 impl ReqwestClient {
     /// Instantiate a new Reqwest client to perform HTTP requests.
-    pub fn new() -> ReqwestClient {
-        Self {
-            reqwest_client: reqwest::Client::new(),
```
ncloudioj, line 25: Just noticed that `request::Client::new()` could panic in certain cases, but the builders let you capture those errors without panicking.

merino-adm/src/remote_settings/reqwest_client.rs
```diff
@@ -20,10 +20,13 @@ pub struct ReqwestClient {
...
-        }
+    pub fn try_new() -> Result<ReqwestClient, Error> {
+        let reqwest_client = reqwest::Client::builder()
+            // Disable the connection pool to avoid the IncompleteMessage errors.
+            // See #259 for more details.
+            .pool_max_idle_per_host(0)
```
Dexterp37, line 27: I'm wondering why this wasn't happening with the other implementation: we were not setting the max idle per host there as well. Any idea?

Aside from that, I don't think there's a big problem in disabling pooling.

mythmon (approved):
I think a better thing to do would be to change the time that `reqwest` keeps connections around so it doesn't outlive the Kinto connections, but I can't find what that value would be. After that, I think adding retry is going to be an important long term solution for many parts of Merino. Neither of these are quick though.

This unblocks us now, and is simple enough to understand. Lets file an issue to come back to this, and merge this fix.

pjenvey:
I'll note that [hyper v0.14.3](https://github.com/hyperium/hyper/blob/master/CHANGELOG.md#v01413-2021-09-16) has a fix around connection reuse + flushing but it definitely does not help, I tried it as a long shot [here](https://app.circleci.com/pipelines/github/mozilla-services/merino/706/workflows/7d89d598-3611-4a1d-979b-bb0c2b660cef/jobs/3209)

pjenvey (approved):

PR #282, opened 2021-12-21
refactor: use actix-web to parse Accept-Language header

ethowitz:
Replaces our home-grown `Accept-Language` parsing logic with Actix Web's, closing #90. I didn't want to introduce Actix Web as a depedency of `merino-suggest`, but I couldn't think of a better way to use the necessary types (`QualityItem`, `Preference`, etc.). Re-exporting them from the `merino-web` crate would require `merino-suggest` to add `merino-web` as a dependency, which creates a cyclic dependency.

### DIFF
--- a/merino-suggest/src/lib.rs
+++ b/merino-suggest/src/lib.rs
@@ -13,7 +13,7 @@ use std::ops::Range;
 use std::time::Duration;
 
 use crate::device_info::DeviceInfo;
-use actix_web::http::header::{AcceptLanguage, LanguageTag, Preference};
+use actix_web::http::header::{AcceptLanguage, LanguageTag, Preference, QualityItem};
 use async_trait::async_trait;
 use fake::{
     faker::{
@@ -340,25 +340,15 @@ pub enum SuggestError {
 pub struct SupportedLanguages(pub AcceptLanguage);
 
 impl SupportedLanguages {
+    /// Returns an instance of [`SupportedLanguages`] that includes every language.
+    pub fn wildcard() -> Self {
+        Self(AcceptLanguage(vec![QualityItem::max(Preference::Any)]))
+    }
+
     /// Specify whether `self` includes the language specified by the given language and region.
-    pub fn includes(&self, language_query: &str, region_query: Option<&str>) -> bool {
-        let language_tag_to_match = {
-            let language_tag = LanguageTag::parse(
-                match region_query {
-                    Some(region) => format!("{}-{}", language_query, region),
-                    None => language_query.to_owned(),
-                }
-                .as_str(),
-            );
-
-            match language_tag {
-                Ok(tag) => tag,
-                Err(_) => return false,
-            }
-        };
-
+    pub fn includes(&self, language_tag: LanguageTag) -> bool {
         self.0.iter().any(|quality_item| {
-            language_tag_to_match.matches(&match &quality_item.item {
+            language_tag.matches(&match &quality_item.item {
                 Preference::Any => return true,
                 Preference::Specific(item) => item.to_owned(),
             })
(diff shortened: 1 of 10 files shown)


merino-suggest/src/lib.rs
```diff
@@ -336,110 +337,46 @@ pub enum SuggestError {
...
     pub fn includes(&self, language_query: &str, region_query: Option<&str>) -> bool {
-        let region_matches = |supported_region| {
-            match (supported_region, region_query) {
-                // If the region query is None, the caller intends to match every region
-                (_, None) => true,
-                // If the region query is Some(_) but the supported region is None, the regions
-                // don't match
-                (None, Some(_)) => false,
-                (Some(supported_region), Some(region_query)) => supported_region == region_query,
+        let language_tag_to_match = {
```
ethowitz, line 345: Would it be preferable for this function just to accept a single `LanguageTag` as its argument?

merino-web/src/extractors.rs
```diff
@@ -86,85 +84,29 @@ struct SuggestQuery {
...
-            } else if let Some((language, region)) = locale_or_wildcard.split_once("-") {
-                Language {
-                    language_identifier: LanguageIdentifier::Locale {
-                        language: language.to_lowercase(),
-                        region: Some(region.to_lowercase()),
-                    },
-                    quality_value,
+        future::ready({
+            if req.headers().contains_key("Accept-Language") {
+                match AcceptLanguage::parse(req) {
```
ethowitz, line 93: `AcceptLanguage::parse` just returns `AcceptLanguage([])` for most types of invalid headers, but it returns the same thing when the request doesn't have an `Accept-Language` header. I handle that by explicitly checking if the header is present and handling the case separately.

merino-web/src/middleware/metrics.rs
```diff
@@ -23,13 +24,14 @@ pub struct MetricsMiddleware<S> {
...
+impl<S, B> Transform<S, ServiceRequest> for Metrics
 where
-    S: Service<ServiceRequest, Response = ServiceResponse>,
+    S: Service<ServiceRequest, Response = ServiceResponse<B>>,
+    B: 'static + MessageBody,
     S::Future: 'static,
     S::Error: fmt::Debug,
 {
-    type Response = ServiceResponse;
+    type Response = S::Response;
```
ethowitz, line 34: I had to make these changes to fix a compile error introduced by the crate upgrades, but I don't fully understand why this solution compiles (or if it's even correct)...I'd appreciate an extra set of 👀 

Same thing for the other `Transform` `impl`s I changed in this PR

merino-suggest/src/lib.rs
mythmon: I think so, yes. That better exposes the error case of if `language_query` and `region_query` are invalid. That better follows the "parse don't verify" pattern.

merino-web/src/extractors.rs
```diff
@@ -86,85 +84,29 @@ struct SuggestQuery {
...
+                    Ok(languages) if languages.is_empty() => {
+                        // AcceptLanguage::parse() returns an empty Vec for certain types of
+                        // errors in the header. In these cases, we want to return an error.
+                        Err(HandlerErrorKind::MalformedHeader("Accept-Language").into())
+                    }
+                    Err(_) => Err(HandlerErrorKind::MalformedHeader("Accept-Language").into()),
```
mythmon, line 99: Returning an error from this function will abort the whole request with a 400 status code. Is that what we want here? I think it may be better to act as if the header wasn't there, and return `QualityItem::max(Preference::Any)`.

I don't think the [RFC for this header](https://httpwg.org/specs/rfc7231.html#rfc.section.5.3.5) specifies a behavior if the header is invalid, but it does include the following:

> If the header field is present in a request and none of the available representations for the response have a matching language tag, the origin server can either disregard the header field by treating the response as if it is not subject to content negotiation or honor the header field by sending a 406 (Not Acceptable) response. However, the latter is not encouraged, as doing so can prevent users from accessing content that they might be able to use (with translation software, for example).

that implies to me that we should err on the side of returning content instead of returning errors.

mythmon (changes requested):

merino-web/src/extractors.rs
ethowitz: Ah yeah, I agree. I'll make the change 👍🏻

mythmon (approved):

Issue #289, opened 2022-01-05
Investigate the memory leak in Merino memory cache


data-sync-user:
The in-memory cache of Merino has shown some symptoms of memory leak in production. In particular, the storage cache layer, as illustrated in the screenshot, shows a constant stepwise increase in the cache size. Whereas the pointer layer shows that the cache expiry job is working periodically on removing lots of the expired pointers. However, it doesn't reclaim too many from the storage layer. Hence we suspect that there are orphaned entries (i.e. entries without any reference pointers associated) in the storage layer.

Note that we haven't fully identified the cause of this behavior, nor has it caused significant concerns in production. Hence it shouldn't be categorized as a high-impact bug for now.

![image](https://user-images.githubusercontent.com/305049/148306664-e0a0ab8a-c513-41fa-89d3-2107c969dd8b.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1507)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/secure/attachment/18992/Screen+Shot+2022-01-05+at+3.08.04+PM.png">Screen Shot 2022-01-05 at 3.08.04 PM.png</a> | <a href="https://mozilla-hub.atlassian.net/secure/attachment/19312/Screen+Shot+2022-01-13+at+1.23.01+PM.png">Screen Shot 2022-01-13 at 1.23.01 PM.png</a>
┆epic: Merino production v1

data-sync-user:
➤ Michael Cooper commented:

I think I found a cause for this, and have filed a PR. We’ll want to leave this issue open until we verify that the problem that PR fixed was causing this issue.

data-sync-user:
➤ Nan Jiang commented:

!Screen Shot 2022-01-13 at 1.23.01 PM.png|width=1244,height=276!

As shown in the screenshot, the memory usage of the storage layer (on the right) of v0.5.1 has largely improved. No signs of constant increase observed over the last few days. It only stores about 10 entries as minimal, which are likely to be the valid ones.

I think the memory leak has been fixed and we can close this now.

PR #290, opened 2022-01-06
fix(deduped-map): Correctly remove pointer and storage entries when calling DedupedMap::remove

mythmon:
This was likely causing #289, since it causes some storage entries to continue
being stored past when their last pointer was removed, due to the reference
count not being updated correctly.

The result of this is that storage entries would leak if their last pointer was
removed via `::remove` instead of via `::retain`.

### DIFF
--- a/deduped-dashmap/src/lib.rs
+++ b/deduped-dashmap/src/lib.rs
@@ -40,6 +40,15 @@ struct MapPointer<M: Debug> {
     hash: u64,
 }
 
+impl<M: Clone + Debug> Clone for MapPointer<M> {
+    fn clone(&self) -> Self {
+        Self {
+            meta: self.meta.clone(),
+            hash: self.hash,
+        }
+    }
+}
+
 /// The second layer of the map, a reference counted value.
 #[derive(Debug)]
 struct MapValue<V> {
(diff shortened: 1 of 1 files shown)


deduped-dashmap/src/lib.rs
```diff
@@ -98,15 +122,15 @@ where
...
+        if let Entry::Occupied(occupied_pointer_entry) = self.pointers.entry(key) {
+            let pointer = occupied_pointer_entry.remove();
             match self.storage.entry(pointer.hash) {
                 Entry::Occupied(mut occupied_storage_entry) => {
                     let item = occupied_storage_entry.get_mut();
                     if item.refcount > 1 {
                         item.refcount -= 1;
                     } else {
-                        occupied_pointer_entry.remove();
+                        occupied_storage_entry.remove();
```
ncloudioj, line 133: Ah, nice catch. The pointer is not removed if the value's refcount is greater than 1. Otherwise, the value will not be removed, which leaves it as an orphan.

deduped-dashmap/src/lib.rs
```diff
@@ -115,6 +139,7 @@ where
                         key = %key_desc, "Dangling storage entry");
                 }
             }
+        } else {
```
ncloudioj, line 142: Why having a blank branch here? Did you mean to add anything here?

deduped-dashmap/src/lib.rs
```diff
@@ -264,9 +289,86 @@ pub enum ControlFlow<C> {
...
+            assert_eq!(pointers[idx].0, day);
+            assert_eq!(pointers[idx].1.meta, day_num);
+            assert_eq!(
+                storage[&pointers[idx].1.hash],
+                MapValue {
+                    value: day_type,
+                    refcount,
+                }
+            );
+        }
```
ncloudioj, line 370: Nit: can you also remove the rest of the items and check the invariant for the sake of completeness?

ncloudioj (approved):

deduped-dashmap/src/lib.rs
mythmon: Oops, left over debugging.

mythmon:
rebased on main

deduped-dashmap/src/lib.rs
mythmon: Is the change I made what you meant?

ncloudioj: Yep, looks good to me, thanks!

ncloudioj (approved):
🚀

PR #291, opened 2022-01-06
refactor(memory-cache): Make the background expiry task a selfish method by adding more Arcs

mythmon:
This makes the function a bit easier to reason about, and requires less special case comments and cloning.

### DIFF
--- a/merino-cache/src/memory.rs
+++ b/merino-cache/src/memory.rs
@@ -17,8 +17,6 @@ use merino_suggest::{
 };
 use std::{
     collections::HashMap,
-    fmt::Debug,
-    hash::Hash,
     sync::Arc,
     time::{Duration, Instant},
 };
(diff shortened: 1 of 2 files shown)


merino-cache/src/memory.rs
```diff
@@ -106,13 +108,18 @@ impl Suggester {
         provider: Box<dyn SuggestionProvider>,
         metrics_client: StatsdClient,
     ) -> Box<Self> {
-        let items = Arc::new(DedupedMap::new());
+        let this = Self {
```
mythmon, line 111: I'm very open to better names that `this` and `that` for the two copies of the suggester object. Note `self` is a reserved word, and cannot be used here.

ncloudioj: Maybe use `suggester` and `cloned`?

ncloudioj (approved):

PR #292, opened 2022-01-07
refactor(memory-cache): Offload the cache expiry task to the blocking threads

ncloudioj:
While Merino uses the "max-removal-entries" to throttle the cache expiry job, it will still prevent other tasks, scheduled on the same thread, from running because those lock acquisitions of the underlying cache stores are inherently blocking. As a result, it might increase the response latency for those affected requests.

Cache cleanup is not of the highest priority and can be done in the background, offloading the expiry task to the blocking threads of the runtime could help smooth those latency spikes.

### DIFF
--- a/merino-cache/src/memory.rs
+++ b/merino-cache/src/memory.rs
@@ -139,7 +139,8 @@ impl Suggester {
                     // is allowed to run.
                     let _ = tokio::task::spawn_blocking(move || {
                         cache.remove_expired_entries();
-                    }).await;
+                    })
+                    .await;
                 }
             });
         }
--- a/merino-cache/src/memory.rs
+++ b/merino-cache/src/memory.rs
@@ -128,14 +128,14 @@ impl Suggester {
                 timer.tick().await;
                 loop {
                     timer.tick().await;
-                    let mut suggester = cloned_suggester.clone();
+                    let mut cache = cloned_suggester.clone();
 
                     // Dispatch the expiry task to the blocking threads of the
                     // runtime. This prevents the expiry task, which is inherently
                     // blocking, from blocking the other tasks running on the
                     // core threads of the runtime.
                     tokio::task::spawn_blocking(move || {
-                        suggester.remove_expired_entries();
+                        cache.remove_expired_entries();
                     });
                 }
             });
--- a/merino-cache/src/memory.rs
+++ b/merino-cache/src/memory.rs
@@ -134,9 +134,12 @@ impl Suggester {
                     // runtime. This prevents the expiry task, which is inherently
                     // blocking, from blocking the other tasks running on the
                     // core threads of the runtime.
-                    tokio::task::spawn_blocking(move || {
+                    //
+                    // Wait for the task to finish so that only one expiry task
+                    // is allowed to run.
+                    let _ = tokio::task::spawn_blocking(move || {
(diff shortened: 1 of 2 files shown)


mythmon:
This can, in theory, change the behavior of the loop. Previously it would wait until the previous task was finished before running the next one, even if the timer would have ticked in the meantime. Under the new system it could have multiple cache expiry tasks overlap if it takes longer to run than the frequency that we run those tasks at.

I'm not sure this is a problem though. What do you think?

 If we did want to avoid that, we would have to do something like add a oneshot channel that we can use to signal the completion of the task.

pjenvey:
> Under the new system it could have multiple cache expiry tasks overlap if it takes longer to run than the frequency that we run those tasks at.

The loop can await the task to serialize these expiry calls

pjenvey:
> The loop can await the spawn_blocking task to serialize these expiry calls

Though I'd also consider moving the loop entirely into the spawn_blocking closure as there's not much point in bouncing back and forth as we loop.

ncloudioj:
> Though I'd also consider moving the loop entirely into the spawn_blocking closure as there's not much point in bouncing back and forth as we loop.

Good point, though I can't find a way to tick the timer inside the closure of spawn_blocking.

pjenvey:
> Good point, though I can't find a way to tick the timer inside the closure of spawn_blocking.

Since it's sync you'd need std::thread::sleep instead.

merino-cache/src/memory.rs
```diff
@@ -128,7 +128,18 @@ impl Suggester {
...
+                    // Dispatch the expiry task to the blocking threads of the
+                    // runtime. This prevents the expiry task, which is inherently
+                    // blocking, from blocking the other tasks running on the
+                    // core threads of the runtime.
+                    //
+                    // Wait for the task to finish so that only one expiry task
+                    // is allowed to run.
+                    let _ = tokio::task::spawn_blocking(move || {
+                        cache.remove_expired_entries();
+                    }).await;
```
ncloudioj, line 142: Looks like await the blocking task (assuming this `JoinHandle` is not blocking) should avoid the repetitive task scheduling concern.

ncloudioj:
> Since it's sync you'd need std::thread::sleep instead.

Would that be a good idea though? Given that the runtime only has a limited number of blocking threads in the pool.

pjenvey:
> Would that be a good idea though? Given that the runtime only has a limited number of blocking threads in the pool.

The pool's default max size is pretty large (512) but you're right in that it's probably not the appropriate place to put a long running "daemon" like thread like this one. We could spawn its own dedicated thread instead.

This isn't terribly important though and the loop calling `spawn_blocking` as you have it is ok.

pjenvey (approved):

ncloudioj:
> The pool's default max size is pretty large (512) but you're right in that it's probably not the appropriate place to put a long running "daemon" like thread like this one. We could spawn its own dedicated thread instead.

Wow, up to 512 threads in the pool. TIL, thanks for that tip!

Yeah, we might have many of those in-memory caches for multiple suggest providers, keeping love-lived daemons for each of those seems like a bit overkill.

Let's see how this works in practice, also we will measure the duration for those expiry tasks. We can move those to a dedicated daemon thread if necessary.

PR #312, opened 2022-01-11
Remove compression features from actix-web

mythmon:
refactor(web): Disable compression related features in actix-web

Nginx handles compression for us, and these features (especially zstd) add
significant compile time. On my machine a clean release compilation went from
90 seconds with zstd to 65 seconds without.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "actix-codec"
-version = "0.4.1"
+version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


pjenvey (commented):
The frontend already compresses responses for us though:

https://github.com/mozilla-services/cloudops-infra/blob/d415c6/projects/merino/k8s/charts/merino/templates/nginx-configmap.yaml#L56

mythmon:
Ah, good to know. The zstd changes are still great for compile times, but this means we can disable even more.

mythmon:
Updated and squashed to just disable the features.

ncloudioj (approved):
🎉

PR #313, opened 2022-01-12
feat(admin): Initial work for dynamic provider configuration

### LINKED ISSUE #296: SuggestionProvider configuration models
This issue covers creating Django models that represent that configuration option of the Merino suggestion providers. These models should be editable in the Django Admin, but don’t need to be exportable to any format.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1520)
┆epic: Merino settings admin: External settings management tool

mythmon:
Fixes CONSVC-1520
Fixes #296


merino-admin/suggestion_providers/models.py
```diff
@@ -0,0 +1,99 @@
...
+        return self.get_real_instance().get_type()
+
+    def get_type(self):
+        return '<unknown>'
+
+
+class RemoteSettingsConfig(ProviderConfig):
+    bucket: models.CharField = models.CharField(max_length=256, blank=True, null=True)
+    collection: models.CharField = models.CharField(max_length=256, blank=True, null=True)
+    resync_interval: models.DurationField = models.DurationField(blank=True, null=True)
```
pjenvey, line 23: These duration fields are currently expected as seconds (field name `resync_interval_sec`) how will the django durations serialize into them?

merino-admin/suggestion_providers/models.py
```diff
@@ -0,0 +1,99 @@
...
+
+class MultiplexerConfig(ProviderConfig):
+    providers: models.ManyToManyField = models.ManyToManyField(ProviderConfig, related_name='multiplexer')
+
+    def get_type(self):
+        return "multiplexer"
+
+
+class TimeoutConfig(ProviderConfig):
+    default_ttl: models.DurationField = models.DurationField()
```
pjenvey, line 58: `TimeoutConfig`'s docs (in ops.md) lack an entry for this and it looks like we misnamed it `default_lock_timeout_sec` anyway? https://github.com/mozilla-services/merino/blob/6cd9f3d/merino-settings/src/providers.rs#L163

merino-admin/suggestion_providers/models.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def provider_type(self):
+        return self.get_real_instance().get_type()
+
+    def get_type(self):
+        return '<unknown>'
+
+
+class RemoteSettingsConfig(ProviderConfig):
+    bucket: models.CharField = models.CharField(max_length=256, blank=True, null=True)
```
ncloudioj, line 21: To make things clearer on the admin page, I'd recommend setting default values for these fields and making them unnullable. What do you think?

merino-admin/suggestion_providers/models.py
ncloudioj: According to the [doc](https://docs.djangoproject.com/en/4.0/ref/models/fields/#durationfield), it's tied to the database of the choice. For SQLite, it's a bigint of microseconds.

In terms of field naming, @mythmon, did you want to stick to the names used in the configuration file?

merino-admin/suggestion_providers/models.py
```diff
@@ -0,0 +1,99 @@
...
+
+class FixedConfig(ProviderConfig):
+    value: models.CharField = models.CharField(max_length=256, )
+
+    def get_type(self):
+        return "fixed"
+
+
+class KeywordFilterConfig(ProviderConfig):
+    suggestion_blocklist: models.TextField = models.TextField()
```
ncloudioj, line 73: This field is of type: HashMap<String, String> in Rust. Perhaps use the JSONField instead? Or just create a model for the blocklist.

merino-admin/suggestion_providers/models.py
mythmon: The database representation isn't important here, it's a storage implementation detail. The Python code will represent these as `timedelta` objects, which are similar to Rust's `Duration` type. When this gets serialized to a format for Merino we can choose how it gets formatted.

merino-admin/suggestion_providers/models.py
mythmon: Bad copy and paste, I'll fix this.

merino-admin/suggestion_providers/models.py
mythmon: I don't think this is the right.

* The type in Merino for this field is `Option<String>`, so it makes sense to allow it to be left unspecified here.
* `blank=True` means that the user is allowed to leave the field unspecified.
* `null=True` means that when a blank value is encountered, it is stored as a `NULL` in the database.

Some help text in the admin could go a long way to fixing this, but I think if we have a default value we'd have to erase it during serialization, which sounds tricky.

ncloudioj: My only concern is that when people add/edit entries on the admin page and see those blank fields, they might get confused about whether or not they should fill them out.

I agree with you that we can also leverage the help text (tooltips?) on the page to advertise we have defaults if unspecified. That works for me, too.

mythmon: Here's what it looks like with the help text I added throughout:

image

merino-admin/suggestion_providers/models.py
```diff
@@ -0,0 +1,330 @@
...
+        help_text="The value to use for the fixed title of the suggestion.",
+    )
+
+    def get_type(self):
+        return "fixed"
+
+
+class Keyword(models.Model):
+    help_text = """
+        A keyword that can be used to match suggestions for filtering.
```
ncloudioj, line 237: Nit: the keyword filters will be matched against the title of the suggestion, not the suggestion per se.

ncloudioj (approved):

pjenvey (approved):

pjenvey (approved):

mythmon:
thanks for the review!

Issue #295, opened 2022-01-07
Periodically reload SuggestionProvider config while the system is running


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1521)
┆epic: Merino settings admin: Load admin settings from a remote system

data-sync-user:
➤ Nan Jiang commented:

This work is now deprioritized. We will revisit this after the Merino tech-stack re-evaludation.

Issue #294, opened 2022-01-07
Fetch SuggestionProvider config from a URL provided in the system settings


data-sync-user:
The static settings for Merino should list a URL that the suggestion provider dynamic settings can be loaded from. This URL should be fetched during the application startup and used to initialize the suggestion providers.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1522)
┆epic: Merino settings admin: Load admin settings from a remote system

mythmon:
This as fixed as a part of #315.

PR #315, opened 2022-01-14
refactor(settings): Split suggestion provider configuration from main settings

### LINKED ISSUE #293: Split SuggestionProvider configuration from main settings
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1523)
┆epic: Merino settings admin: Load admin settings from a remote system

ncloudioj:
This fixes #293 

Change highlights:
* Moved the configurations for suggestion provider to its own settings
* Added a new helper struct `SuggestionProviderSettings` to load settings. My understanding is that we can build up the dynamic loading around it later.   

Notes:
* This requires a change to `cloudops-infra`       
* The configuration sources for providers and the overriding rules are identical to the top-level settings. I am open to further simplifying that if necessary
* The top-level `Settings` still hold the `suggestion_providers` as I don't see much benefit of spitting it from the `Settings` for now.

### DIFF
--- a/merino-settings/Cargo.toml
+++ b/merino-settings/Cargo.toml
@@ -5,9 +5,11 @@ edition = "2018"
 
 [dependencies]
 anyhow = "1.0"
-config = "0.11.0"
+async-trait = "0.1"
+config = { git = "https://github.com/mehcode/config-rs", branch = "master" }
 http = "0.2"
 redis = "^0.20"
+reqwest = "0.11"
 # Pin to 0.19 until our on premise server updates to >= 20.6.
 sentry = "0.19"
 serde = { version = "1.0.125", features = ["derive"] }
@@ -17,5 +19,7 @@ tracing = "0.1.29"
 tracing-subscriber = { version = "0.2.18", features = ["env-filter"] }
 
 [dev_dependencies]
+httpmock = "0.6"
 parameterized = "0.3"
 serde_json = "1"
+tokio = { version = "1", features = ["macros", "rt"] }
--- a/merino-settings/Cargo.toml
+++ b/merino-settings/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2018"
 [dependencies]
 anyhow = "1.0"
 async-trait = "0.1"
-config = { git = "https://github.com/mehcode/config-rs", branch = "master" }
+config = { git = "https://github.com/mehcode/config-rs", rev = "53e43fb" }
 http = "0.2"
 redis = "^0.20"
 reqwest = "0.11"
--- a/merino-settings/src/providers.rs
+++ b/merino-settings/src/providers.rs
@@ -1,5 +1,10 @@
+use crate::ProviderSettings;
 use anyhow::{Context, Result};
-use config::{Config, File};
+use async_trait::async_trait;
+use config::{
+    builder::AsyncState, AsyncSource, Config, ConfigBuilder, ConfigError, File, FileFormat, Format,
+    Value,
+};
 use serde::{Deserialize, Serialize};
 use serde_with::{serde_as, DurationMilliSeconds, DurationSeconds};
 use std::collections::HashMap;
@@ -226,7 +231,9 @@ impl SuggestionProviderSettings {
     /// Load settings for suggestions providers.
     ///
     /// The organization of the provider configuration files is identical to the
-    /// top level settings.
+    /// top level settings except that it only uses one source (local or remote)
+    /// for each run-mode, and that is configured by `provider_settings` of the
+    /// top-level settings.
     ///
     /// Note that settings for suggestion providers cannot be configured via
     /// environment variables.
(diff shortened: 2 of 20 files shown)


merino-settings/src/providers.rs
```diff
@@ -215,6 +217,60 @@ impl Default for StealthConfig {
     }
 }
 
+/// Settings for Merino suggestion providers.
+#[derive(Serialize, Deserialize)]
+pub struct SuggestionProviderSettings {
+    pub suggestion_providers: HashMap<String, SuggestionProviderConfig>,
```
ncloudioj, line 223: This requires each config file to have a top-level "suggestion_providers", which seems redundant. I've tried defining `struct SuggestionProviderSettings(pub HashMap<String, SuggestionProviderConfig>)`, but couldn't get the deserialization working.

mythmon:
This isn't quite what I had in mind, but that's my fault since I didn't actually explain what I was thinking. I was expecting to have an entirely separate source of configuration, that wasn't related to environment variables or the configuration library. If this way works though, I don't have any hard objections.

Does this seem like a helpful first step to progressing down the road of dynamic settings? Maybe it would be better to implement the next issue in this series on top of this, to see if any changes need to be made here.

merino-settings/src/providers.rs
mythmon: I'm not having trouble with tuple-types wrapping hash maps. Can you recreate the problem in a smaller case?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9d374e3390338ae95239835c11811ca5

ncloudioj:
> This isn't quite what I had in mind, but that's my fault since I didn't actually explain what I was thinking.

No worries, I was also not sure about what your original plan was, so just wanted to take a quick stab for discussions. Totally Ok to toss this away for a rework :)

> I was expecting to have an entirely separate source of configuration, that wasn't related to environment variables or the configuration library. If this way works though, I don't have any hard objections.

Are you expecting to completely rely on the dynamic settings without having a local configuration as a fallback? Like fetching settings from remote at startup and then periodically pulling (or waiting for push) during runtime.

ncloudioj:
> Does this seem like a helpful first step to progressing down the road of dynamic settings? Maybe it would be better to implement the next issue in this series on top of this, to see if any changes need to be made here.

Taking a closer look at #294 and #295, I think I understand what the game plan is now :).

My only question is: shall we consider having a local fallback? It might be out-of-dated, but at least can be used in case of the remote one is unavailable.

ncloudioj:
Looks like that configuration library also supports the [async source](https://github.com/mehcode/config-rs/blob/master/examples/async_source/main.rs) out of the box.

merino-settings/src/providers.rs
ncloudioj: Yeah, that worked for me too. But when I changed this to

```rs
#[derive(Serialize, Deserialize)]
pub struct SuggestionProviderSettings (
    pub HashMap<String, SuggestionProviderConfig>
);
```

Setting `config/providers/test.yaml` to `{}` or a valid object. It yelled at me with this:

```
'Could not convert settings for suggestion providers: invalid type: map, expected tuple struct SuggestionProviderSettings', merino-settings/src/providers.rs:270:14
```

merino-settings/src/providers.rs
```diff
@@ -215,6 +217,60 @@ impl Default for StealthConfig {
...
+        let merino_env = std::env::var("MERINO_ENV").unwrap_or_else(|_| "development".to_string());
+        s.merge(File::with_name(&format!("./config/providers/{}", merino_env)).required(false))
+            .context("loading environment config for suggestion providers")?;
+
+        // Add a local configuration file that is `.gitignore`ed.
+        s.merge(File::with_name("./config/providers/local").required(false))
+            .context("loading local config for suggestion providers")?;
+
+        serde_path_to_error::deserialize(s)
+            .map(SuggestionProviderSettings)
```
ncloudioj, line 253: This seems to work around the deserialization error, though I am still not entirely sure why the direct deserialization didn't work in the first place.

merino-settings/src/providers.rs
ncloudioj: Turns out I have to explicitly enable the [`transparent`](https://serde.rs/container-attrs.html#transparent) attribution to make it work.

merino-settings/src/providers.rs
```diff
@@ -215,6 +217,59 @@ impl Default for StealthConfig {
...
+
+        serde_path_to_error::deserialize(s)
+            .context("Deserializing settings for suggestion providers")
+    }
+
+    /// Load settings for suggestion providers from configuration files for tests.
+    pub fn load_for_tests() -> Self {
+        let mut s = Config::new();
+
+        s.merge(File::with_name("../config/providers/test"))
```
pjenvey, line 261: Though there's nothing in there currently I think this should also load `../config/providers/base` first similar to what `Settings` does.

pjenvey (changes requested):

merino-settings/src/providers.rs
pjenvey: Or maybe this doesn't matter since this is solely a `HashMap`? I'm not sure how merge/try_into work here..

pjenvey: Testing locally the merging does support this, e.g. a `test_remote_settings` in `test.yaml` and `test_wiki_fruit` in `local_test.yaml` results in both keys included. So it makes sense to also load from `base`.

mythmon:
> My only question is: shall we consider having a local fallback? It might be out-of-date, but at least can be used in case of the remote one is unavailable. The local setting could be useful for other non-production environments as well.

Especially with the configuration library supporting async sources, I'm liking this setup more and more. I am constantly surprised by the flexibility of that library.

I think that the next step here is to make the path that the provider configs are loaded from configurable. The default value (especially for local dev) can be a `file://` URI that points to a in-repo sampple. In production we'd use an `https://` URI that loads the real config from Shepherd.

I'm not convinced we'd need multiple layers so that it can fallback though. In production a failure to load the config at startup should probably cause retries and failure to start.

merino-settings/src/providers.rs
ncloudioj: Yeah, since it's backed by `HashMap`, `merge` will override the old one if there is a new one with the same provider ID.

I think the reason I didn't load from `base` is that some special providers (WikiFruit, DebugProvider, and FixedProvider) are only valid under the debug mode, trying to load them for other modes will fail.

I wonder if we should just get rid of `base` because, unlike the top-level settings, I don't see a common provider that's shared across all the different modes. So for suggestion providers, we will have a scenario-specific configuration (e.g. dev, prod, test, ci, etc), a local one (optional & gitignored), and another remote one for production.

@pjenvey @mythmon What do you all think?

mythmon: I think you're right that there aren't currently any common providers we'd want to share across providers. Maybe that will change in the future, but if it does then we can add this back. Also as I explained elsewhere, I think we should only support a single configuration source for this anyways.

ncloudioj: Having a single source and loading the source path/URL from the top-level settings sounds like a good idea to me. As for those local sources (`providers/local.yaml` & `providers/local_test.yaml`), I think folks could still use them by overriding the source path in their `local.yaml`. Then all the major scenarios should be covered by this.

pjenvey: Works for me

pjenvey:
Woops, I missed the discussion from last week. The async source is a nice find, I didn't know about it probably because it's omitted from their rustdocs.

cc: @jrconlin -- who implemented similar functionality for contile but it reads from a Cloud Storage Bucket (from a `gs://` instead of a plain url). I think the potential added benefit there is the ability to secure the settings file behind GCP's auth.

pjenvey (approved):

merino-settings/Cargo.toml
```diff
@@ -5,9 +5,11 @@ edition = "2018"
 
 [dependencies]
 anyhow = "1.0"
-config = "0.11.0"
+async-trait = "0.1"
+config = { git = "https://github.com/mehcode/config-rs", branch = "master" }
```
pjenvey, line 9: Looks great but I think we should pin this in case their master breaks on us later

```suggestion
config = { git = "https://github.com/mehcode/config-rs", rev = "53e43fb" }
```

ncloudioj: Good point!

ncloudioj:
Sigh, there is one more change with `config-rs` regarding how it parses the environment variables. Specifically, once we specify a custom separator (e.g. `__`), it also applies it all the way up to the prefix (e.g. `MERINO`) rather than assuming the top-level separator is a single `_`. So now, `MERINO_HTTP__URL` won't work, it only takes `MERINO__HTTP__URL`.

This change is kind of unwelcoming as we have been used to the old convention, but it appears to be more consistent. I think we can get used to it.

This will require a change to Merino's production configuration.

mythmon:
> This will require a change to Merino's production configuration.

This is all going to be a big breaking change, so I think this is ok. I've been thinking about tagging 1.0 soon anyways.

mythmon (approved):

pjenvey (approved):

ncloudioj:
Thank you all for the review!

PR #317, opened 2022-01-19
chore: Update to edition 2021

mythmon:


Dexterp37 (changes requested):
Whoopsy,looks like the docker image doesn't like this :( Would you kindly fix that?

Dexterp37 (approved):

PR #323, opened 2022-01-27
feat: add client_aware provider [CONSVC-1423]

### LINKED ISSUE #271: Add client_variant aware provider
Merino's API includes a client_variants field as a way for Firefox to inform Merino about experiments that are running on the client. Merino should be able to modify its behavior based on that field.

To this end, Merino should have a provider combinator that selects between multiple subproviders based on the provided client variants. One request can have multiple client variants, so likely this should be a simple boolean switch between having a particular variant or not. A more complicated switch-statement-like provider is probably overkill at this time.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1423)

tiftran:
fixes #271


merino-suggest/src/providers/client_variant_filter.rs
```diff
@@ -0,0 +1,178 @@
...
+    fn cache_inputs(&self, request: &SuggestionRequest, cache_inputs: &mut dyn CacheInputs) {
+        self.matching_provider.cache_inputs(request, cache_inputs);
+        self.default_provider.cache_inputs(request, cache_inputs);
+
+        let req = request.clone();
+
+        if req
+            .client_variants
+            .unwrap_or(vec![])
+            .contains(&self.client_variant)
```
mythmon, line 66: I think you could do this without cloning the entire request like:

```suggestion
        if request
            .client_variants
            .clone()
            .unwrap_or(vec![])
            .contains(&self.client_variant)
```

That's a bit of a micro-optimization, but I think it more clearly highlights your intent. (If that doesn't work for some reason, I'm curious about the error it causes).

merino-suggest/src/providers/client_variant_filter.rs
```diff
@@ -0,0 +1,178 @@
...
+        CacheStatus, ClientVariantFilterProvider, SuggestError, Suggestion, SuggestionProvider,
+        SuggestionRequest, SuggestionResponse,
+    };
+    use async_trait::async_trait;
+    use fake::{Fake, Faker};
+
+    struct TestMatchSuggestionsProvider();
+
+    #[async_trait]
+    impl SuggestionProvider for TestMatchSuggestionsProvider {
```
mythmon, line 89: Instead of defining two bespoke providers, it *might* be easier to use the `Fixed` provider. It's a debugging provider where you specify a string in it's config, and it simply returns a suggestion with that string as the title. It's supposed to be for this kind of situation, but it hasn't been used much yet.

merino-web/src/providers.rs
```diff
@@ -98,6 +98,26 @@ async fn make_provider_tree(
...
+            let default_provider = make_provider_tree(
+                settings,
+                filter_config.default_provider.as_ref(),
+                metrics_client,
+            )
+            .await?;
+            ClientVariantFilterProvider::new_boxed(
+                matching_provider,
+                default_provider,
+                (*filter_config.client_variant).to_string(),
```
mythmon, line 117: I think this might be clearer written as
```suggestion
                filter_config.client_variant.clone(),
```

mythmon:
Rebasing this on the latest `main` will fix the audit issue.

merino-suggest/src/providers/client_variant_filter.rs
pjenvey: You can even go a little farther and avoid any cloning here:
```suggestion
        if request
            .client_variants
            .as_ref()
            .map_or(false, |cv| cv.contains(&self.client_variant))
```

tiftran: oooh fancy

merino-suggest/src/providers/client_variant_filter.rs
```diff
@@ -0,0 +1,178 @@
...
+            .contains(&self.client_variant)
+        {
+            &self.matching_provider
+        } else {
+            &self.default_provider
+        };
+        let results = provider
+            .suggest(request)
+            .await
+            .unwrap_or_else(|_| SuggestionResponse::new(vec![]));
```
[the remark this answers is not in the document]
tiftran: not sure...

merino-settings/src/providers.rs
```diff
@@ -222,6 +223,25 @@ impl Default for StealthConfig {
...
+#[serde_as]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(default)]
+pub struct ClientVariantSwitchConfig {
+    pub client_variant: String,
+    pub matching_provider: Box<SuggestionProviderConfig>,
+    pub default_provider: Box<SuggestionProviderConfig>,
+}
+
+impl Default for ClientVariantSwitchConfig {
```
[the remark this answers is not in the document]
tiftran: that makes sense

merino-admin/suggestion_providers/admin.py
```diff
@@ -35,10 +36,13 @@ class BaseConcreteProviderConfigAdmin(PolymorphicChildModelAdmin):
...
-                'fields': self.get_fields(request, obj=obj),
-                'description': self.base_model.help_text
-            })
+            (
+                None,
+                {
+                    "fields": self.get_fields(request, obj=obj),
+                    "description": self.base_model.help_text,
+                },
+            )
```
mythmon, line 45: Huh. I guess I forgot to run Black on this when I made my PR.

merino-admin/suggestion_providers/models.py
```diff
@@ -281,6 +281,41 @@ def get_type(self):
...
+            will be used for suggestions. If not, the default_provider will be 
+            used.
+        """,
+    )
+    matching_provider: models.ForeignKey = models.ForeignKey(
+        ProviderConfig,
+        on_delete=models.SET_NULL,
+        null=True,
+        related_name="client_variant_switch_matching_provider",
+        help_text=INNER_PROVIDER_HELP_TEXT,
```
mythmon, line 304: It might be good to use a specific help text or add something to the constant provider help text to indicate that this provider is used specifically in the matching case. Same with the default one.

merino-settings/src/providers.rs
```diff
@@ -366,6 +375,7 @@ mod tests {
                     | SuggestionProviderConfig::Timeout(_)
                     | SuggestionProviderConfig::KeywordFilter(_)
                     | SuggestionProviderConfig::Stealth(_)
+                    | SuggestionProviderConfig::ClientVariantSwitch(_)
```
mythmon, line 378: You also need to add a sample client variant settings block to the `value_json` variable above, and increment the `found_providers` assertion down below.

merino-suggest/src/providers/client_variant_filter.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    async fn suggest(
+        &self,
+        request: SuggestionRequest,
+    ) -> Result<SuggestionResponse, SuggestError> {
+        let req = request.clone();
+        let provider = if req
+            .client_variants
+            .unwrap_or_default()
+            .contains(&self.client_variant)
```
mythmon, line 49: I realize that you can do the same `as_ref` trick here that you used in `cache_inputs`.

merino-web/src/extractors.rs
```diff
@@ -65,17 +69,23 @@ impl FromRequest for SuggestionRequestWrapper {
...
 /// A query passed to the API.
+#[serde_as]
 #[derive(Debug, Deserialize)]
 struct SuggestQuery {
     /// The query to generate suggestions for.
     q: String,
+    /// The client variants to generate suggestions with.
+    #[serde_as(as = "Option<StringWithSeparator::<CommaSeparator, String>>")]
+    #[serde(default)]
+    client_variants: Option<Vec<String>>,
```
mythmon, line 88: As a future consideration, it might be worth figuring out how to make this just a `Vec<String>`, instead of making an `Option`. I don't think at this level there is a semantic difference between no client variants and an empty list.

mythmon (approved):

PR #333, opened 2022-02-03
feat: Reconfigurable providers

mythmon:
Moving the provider implementations away from the basic traits makes room to have tools like `make_provider_tree` in a common place without creating dependency loops.


merino-web/src/providers.rs
```diff
@@ -3,128 +3,37 @@
...
-    ClientVariantFilterProvider, DebugProvider, FixedProvider, IdMulti, KeywordFilterProvider,
-    Multi, NullProvider, StealthProvider, SuggestionProvider, TimeoutProvider, WikiFruit,
-};
+use merino_settings::Settings;
+use merino_suggest_providers::make_provider_tree;
+use merino_suggest_providers::IdMulti;
 
 /// The SuggestionProvider stored in Actix's app_data.
 #[derive(Clone)]
 pub struct SuggestionProviderRef(pub Arc<IdMulti>);
```
ncloudioj, line 13: Do we need to change this to `SuggestionProviderRef(pub RwLock<IdMulti>)` to support the provider reloading?

Currently, we keep an [instance](https://github.com/mozilla-services/merino/blob/main/merino-web/src/lib.rs#L127) of `SuggestionProviderRef` for each Actix worker. We will need mutable access In order to call `reconfigure()` on those individual providers, since Actix State is already wrapped by Arc, looks like we can wrap it into a RwLock here so that all the Actix works can share a single instance of this struct.

@mythmon Is that what you're planning for provider reloading?

mythmon: When I wrote this I didn't yet have a plan that covered that. In the intervening time, I've updated `SuggestionProviderRef` to be:

```rust
/// The SuggestionProvider stored in Actix's app_data.
#[derive(Clone)]
pub struct SuggestionProviderRef {
    /// The wrapped providers.
    pub provider: Arc<TokioRwLock<IdMulti>>,
    /// The metrics client used, to enable reconfiguration of providers.
    metrics_client: StatsdClient,
    /// The settings used to create the provider, to enable reconfiguration.
    settings: Settings,
}
```

Keeping an extra copy of the settings and metrics clients isn't my favorite, but it makes updates a lot more straight forward. I'm not 100% sure that I want to keep it, but it's working for now so I can make progress.

ncloudioj: Sounds good.

Does it make sense to wrap `IDMulti` inside of an `ArcSwap` here?  it's specifically designed for this read-mostly write-seldom scenario. We might be able to get away from all those locking overheads somehow.

mythmon: Wouldn't `ArcSwap` mean we have to replace the provider with a new instance entirely? That would mean we would always lose state when we reconfigure. I thought the whole point of doing this reconfigure part instead of a simple swap was so we don't lose state, like caches.

ncloudioj: Oh right, we can't do ArcSwap here. Sorry, I totally missed that.

Reconfigure is now covered, can it also support provider removal as well? Like removing an ID'd provider from IdMulti.

ncloudioj: Looks like we can easily add the removal functionality to `IdMulti`. Changing provider type for the same ID also seems doable, though it might not be very useful in practice.

mythmon: Yep, the current version of ID multi supports adding, removing, and updating in place for things where the ID doesn't change.

mythmon:
(meant to post this yesterday, oops)

I've added implementations for reconfigure for most of the providers. I haven't done the remote setting provider yet, since I think it's going to take some major refactoring to accomplish. I've also added a dangerous debug endpoint, `/api/v1/providers/reconfigure` which takes *unvalidated* user input and *unconditionally* replaces the current config with it. This is obviously something that we never want to have enabled in production, but it helps test this workflow manually.

merino-suggest-traits/src/lib.rs
```diff
@@ -360,6 +369,72 @@ impl SupportedLanguages {
...
+/// If the `Value` passed does not conform to the expected schema of the config,
+/// a `SetupError::InvalidConfiguration` error will be returned.
+pub fn convert_config<T: DeserializeOwned>(config: serde_json::Value) -> Result<T, SetupError> {
+    serde_json::from_value::<T>(config)
+        .context("loading provider config")
+        .map_err(SetupError::InvalidConfiguration)
+}
+
+/// Reconfigure the passed provider, or if there is a problem reconfiguring it,
+/// replace it with a newly created provider.
```
ncloudioj, line 414: Can you note here that if `make_fresh` also fails, the original provider will remain unchanged?

ncloudioj (approved):
Looks great, thanks! 42 files changed, you picked a nice number :)

Also filed three issues to track the follow up items.

PR #336, opened 2022-02-11
fix: don't forget to add in the fresh providers

pjenvey:


merino-suggest-providers/src/providers/multi.rs
```diff
@@ -99,7 +99,7 @@ impl SuggestionProvider for Multi {
                 queue.push(make_fresh(conf));
             }
             while let Some(result) = queue.next().await {
-                result?;
+                self.providers.push(result?);
```
Dexterp37, line 102: Can we add test coverage for this?

Dexterp37 (changes requested):

merino-suggest-providers/src/providers/multi.rs
ncloudioj: FWIW, the initial PR (#333) merged yesterday was known for lack of tests, I've filed this [follow up](https://mozilla-hub.atlassian.net/browse/CONSVC-1617) to fix that. But I'll let @pjenvey decide if some basic tests can be added here now.

Dexterp37: Happy to waive on the test request if we feel comfortable for this one, since we have follow-up work already planned.

pjenvey: Let's merge this as is, just as a quick follow up to #333 and get to the tests later.

Dexterp37 (approved):

PR #337, opened 2022-02-15
test: add unit tests for provider reconfigure

ncloudioj:
This fixes [CONSVC-1617](https://mozilla-hub.atlassian.net/browse/CONSVC-1617) sd s follow-up of #333.

Unit tests are added to all re-configurable providers except "adm-provider" and "redis-provider" as the former is not fully implemented and the latter is only testable through integration tests, and I haven't figured out an easy way to do that.

The "wikifruit" and "debug" providers are also not tested as there is no point in re-configuring them.

As an aside, looks like we can further improve the re-configure implementation to make it more ergonomic. For example, having a way to only re-configure the outer provider. Also, we can make the inner provider maker `make_fresh` optional rather than mandatory so that if the re-configure fails on the inner provider(s), simply re-use the old one rather than trying to create a new one.

### DIFF
--- a/merino-cache/src/memory.rs
+++ b/merino-cache/src/memory.rs
@@ -331,7 +331,9 @@ mod tests {
     use cadence::{SpyMetricSink, StatsdClient};
     use deduped_dashmap::DedupedMap;
     use fake::{Fake, Faker};
-    use merino_suggest_traits::{NullProvider, Suggestion};
+    use futures::{future::ready, FutureExt};
+    use merino_settings::providers::{MemoryCacheConfig, SuggestionProviderConfig};
+    use merino_suggest_traits::{MakeFreshType, NullProvider, Suggestion, SuggestionProvider};
     use std::{
         sync::Arc,
         time::{Duration, Instant},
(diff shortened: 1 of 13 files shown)


merino-cache/src/memory.rs
```diff
@@ -413,4 +415,41 @@ mod tests {
...
+            metrics_client,
+            items: cache.clone(),
+            default_ttl: Duration::from_secs(30),
+            default_lock_timeout: Duration::from_secs(1),
+            max_background_removals: usize::MAX,
+        };
+
+        // This won't be called as `DelayProvider::reconfigure()` will always succeed.
+        let make_fresh: MakeFreshType = Box::new(move |_fresh_config: SuggestionProviderConfig| {
+            ready(Ok(Box::new(NullProvider) as Box<dyn SuggestionProvider>)).boxed()
```
Dexterp37, line 437: Since we claim this never gets called, should we put an `assert!(false)` here?

Dexterp37: (if you think it would be useful asserting here, please kindly make sure to assert in the other tests as well)

Dexterp37 (approved):
Thanks for adding tests! r+ with a minimal, optional nit :)

merino-cache/src/memory.rs
ncloudioj: Good point! Thanks!

I ended up adding `unreachable()` as Clippy was picky on `assert!(false)`.

PR #343, opened 2022-02-23
refactor(adm) - Add clear to deduped-dashmap for fast map resetting

ncloudioj:


deduped-dashmap/src/lib.rs
```diff
@@ -235,6 +235,15 @@ where
     pub fn contains_key(&self, key: &K) -> bool {
         self.pointers.contains_key(key)
     }
+
+    /// Clears the map, both the pointer and the storage maps.
+    ///
+    /// It behaves as if `Self::retain(|_, _, _| ControlFlow::Continue(false))`
+    /// except being faster as the predicate checking step is skipped.
+    pub fn clear(&self) {
+        self.storage.clear();
```
pjenvey, line 244: Nit: as we don't lock across these operations and especially because we're more read heavy than write I'd argue clearing the pointers before storage would result in less 'dangling pointer' situations (which we do handle, hence this being a nit, but they emit a log messages). Albeit they will be rare, regardless.

ncloudioj: Good point, will change the order and add a note for this.

I don't think it will cause any trouble as it's only used by `merino-adm`, where the `deduped-dashmap` is always mutated behind the provider `RWLock`. But yeah, if this gets called in other cases such as in `merino-cache`, we might run into that issue.

pjenvey (approved):

PR #344, opened 2022-02-25
feat(adm): Add retry for Remote Settings provider

ncloudioj:
This fixes [CONSVC-1635](https://mozilla-hub.atlassian.net/browse/CONSVC-1635).

Change highlights:
* Added the retry mechanism for Remote Settings resyncing
* Added a new config `cron_interval` to `RemoteSettingsGlobalSettings`, which would require a change in `cloud-infra`
* The resyncing and the retry-on-failures are now done inside of this new "cron" job. It ticks much faster than `resync_interval` so that it will keep retrying until the resync succeeds. I believe that this retry mechanism is more desirable than retry-up-N-times as we'd like Merino to fetch the latest content from RS as soon as possible   
* This also simplified the reconfigure handling, as now it only needs to send the new configuration to this cron job


pjenvey (approved):

ncloudioj:
Thanks for the review!

I will make a PR for `cloudops-infra` to add this configuration, and then merge it.

Issue #352, opened 2022-03-11
Implement the build promotion mechanism (stage -> prod)


data-sync-user:
This ticket is for tracking the implementation of this proposal [https://docs.google.com/document/d/1WmkZGthJPUSgRxdG1Z5w-lZU4IDgdVbqX7lx6Fh-6G4/edit#](https://docs.google.com/document/d/1WmkZGthJPUSgRxdG1Z5w-lZU4IDgdVbqX7lx6Fh-6G4/edit#|smart-link) 

This requires some SRE effort and some Merino devs time (I’ll commit to it on the Merino part).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1670)
┆epic: Investigate and enable continuous deployment for Merino

data-sync-user:
➤ Alessio Placitelli commented:

Dustin Lactin hey! We can finally start with the implementation for this 🙂 Last time we chatted about it you mentioned it would require some changes on the Jenkins end in addition to the changes on the CircleCi part. Could these changes be tackled separately, or do we need to coordinate and ship them at the same time?

I would love to make sure the Merino team is able to still ask SRE to perform emergency manual deployments (in the unlikely event that’s needed) until we get done with this.

Dustin Lactin do you think we should meet and briefly talk about the implementation for this again?

data-sync-user:
➤ Dustin Lactin commented:

PR created for pipeline updates: https://github.com/mozilla-services/cloudops-infra/pull/3915 ( https://github.com/mozilla-services/cloudops-infra/pull/3915|smart-link ) 

(Regex update & removal of input steps from production stages)

PR #353, opened 2022-03-17
consvc-1670 - Enable deploying to production

Dexterp37:
This introduces the CircleCI facilities that are used to trigger deployment to production.

It also changes the release process documents to highlight the new release process.

Here's what the UI looks like (I tried this in my [toy repo](https://github.com/Dexterp37/test_ci)):

image

Works toward, but does not fully fix, #352 (it still lacks the SRE changes on the pipeline side).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,7 +49,7 @@ workflows:
             - checks
             - build-and-test
             - docker-image-build
-      - docker-image-publish:
+      - docker-image-publish-stage:
           <<: *main-filters
           requires:
             - checks
@@ -158,7 +158,7 @@ jobs:
            docker-compose -f test-engineering/contract-tests/docker-compose.yml build client kinto-setup
            docker-compose -f test-engineering/contract-tests/docker-compose.yml up --abort-on-container-exit
 
-  docker-image-publish:
+  docker-image-publish-stage:
     docker:
       - image: mozilla/cidockerbases:docker-latest
         auth:
@@ -179,10 +179,6 @@ jobs:
           command: |
             if [ "${CIRCLE_BRANCH}" == "main" ]; then
               DOCKER_TAG="${CIRCLE_SHA1}"
-            fi
-
-            if [ -n "${CIRCLE_TAG}" ]; then
-              DOCKER_TAG="$CIRCLE_TAG"
             fi
 
             if [ -n "${DOCKER_TAG}" ]; then
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,6 +56,22 @@ workflows:
             - build-and-test
             - docker-image-build
             - contract-tests
+      # The following job will require manual approval in the CircleCI web application.
+      # Once provided, and when all the requirements are fullfilled (e.g. tests)
+      - unhold-to-deploy-to-prod:
+          <<: *main-filters
+          type: approval
+          requires:
+            - checks
+            - build-and-test
+            - docker-image-build
+            - contract-tests
+      # On approval of the `unhold-to-deploy-to-prod` job, any successive job that requires it
+      # will run. In this case, it's manually triggering deployment to production. 
+      - docker-image-publish-prod:
+          <<: *main-filters
+          requires:
+            - unhold-to-deploy-to-prod
 
 jobs:
   checks:
@@ -179,6 +195,38 @@ jobs:
           command: |
             if [ "${CIRCLE_BRANCH}" == "main" ]; then
               DOCKER_TAG="${CIRCLE_SHA1}"
+            fi
+
+            if [ -n "${DOCKER_TAG}" ]; then
+              echo ${DOCKERHUB_REPO}:${DOCKER_TAG}
+              docker tag app:runtime ${DOCKERHUB_REPO}:${DOCKER_TAG}
+              docker images
+              docker push "${DOCKERHUB_REPO}:${DOCKER_TAG}"
+            else
+              echo "Not pushing to dockerhub for tag=${CIRCLE_TAG} branch=${CIRCLE_BRANCH}"
+            fi
+
+  docker-image-publish-prod:
+    docker:
+      - image: mozilla/cidockerbases:docker-latest
+        auth:
+          username: $DOCKER_USER
+          password: $DOCKER_PASS
+    steps:
+      - checkout
+      - skip-if-do-not-deploy
+      - setup_remote_docker
+      - attach_workspace:
+          at: workspace
+      - dockerhub-login
+      - run:
+          name: Load Docker image from workspace
+          command: docker load -i workspace/merino.tar.gz
+      - run:
+          name: Push to Dockerhub (prod)
+          command: |
+            if [ "${CIRCLE_BRANCH}" == "main" ]; then
+              DOCKER_TAG="prod-${CIRCLE_SHA1}"
             fi
 
             if [ -n "${DOCKER_TAG}" ]; then
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -191,7 +191,7 @@ jobs:
           name: Load Docker image from workspace
           command: docker load -i workspace/merino.tar.gz
       - run:
-          name: Push to Dockerhub
+          name: Push to Docker Hub
           command: |
             if [ "${CIRCLE_BRANCH}" == "main" ]; then
               DOCKER_TAG="${CIRCLE_SHA1}"
@@ -223,7 +223,7 @@ jobs:
           name: Load Docker image from workspace
           command: docker load -i workspace/merino.tar.gz
       - run:
-          name: Push to Dockerhub (prod)
+          name: Push to Docker Hub (prod)
           command: |
             if [ "${CIRCLE_BRANCH}" == "main" ]; then
               DOCKER_TAG="prod-${CIRCLE_SHA1}"
--- a/docs/dev/release-process.md
+++ b/docs/dev/release-process.md
@@ -28,9 +28,13 @@ While the `[do not deploy]` can be anywhere in the title, it is recommended to p
 The deployment pipeline will analyse the message of the merge commit (which will be contain the PR title) and make a decision based on it.
 
 ## Releasing to production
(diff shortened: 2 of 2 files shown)


docs/dev/release-process.md
```diff
@@ -28,5 +28,27 @@ While the `[do not deploy]` can be anywhere in the title, it is recommended to p
 The deployment pipeline will analyse the message of the merge commit (which will be contain the PR title) and make a decision based on it.
 
 ## Releasing to production
-The process to promote a build from `stage` to `production` is currently manually initiated by SRE.
-[This ticket](https://mozilla-hub.atlassian.net/browse/CONSVC-1566) (requires LDAP) deals with automating the process.
+Developers with write access to the Merino repository can initiate a deployment to production after a Pull-Request on the Merino GitHub repository is merged to the `main` branch. This can be done by:
```
ncloudioj, line 31: Just for clarification, are `developers` referring to the same individual(s) who authored and merged the pull request, or do we need someone else to operate this including the approval process mentioned below? I think it should be the former but just wanted to double check.

docs/dev/release-process.md
```diff
@@ -28,5 +28,27 @@ While the `[do not deploy]` can be anywhere in the title, it is recommended to p
...
+1. opening the [circleci dashboard][circleci_dashboard];
+2. looking up the pipeline named `merino <PR NUMBER>` running in the `main-workflow`; this pipeline should either be in a running status (if the required test jobs are still running) or in the "on hold" status, with the `unhold-to-deploy-to-prod` being held;
+3. once in the "on hold" status, with all the other jobs successfully completed, clicking on the "thumbs up" action on the `unhold-to-deploy-to-prod` job row will approve it and trigger the deployment, unblocking the `deploy-to-prod` job;
+4. developers **must** monitor the [Merino Application & Infrastructure][merino_app_info] dashboard for any anomaly, for example significant changes in HTTP response codes, increase in latency, cpu/memory usage (most things under the infrastructure heading).
+
+[circleci_dashboard]: https://app.circleci.com/pipelines/github/mozilla-services/merino?branch=main&filter=all
+[merino_app_info]: https://earthangel-b40313e5.influxcloud.net/d/Cm83vS57z/merino-application-and-infrastructure?orgId=1&refresh=1m
+
+## What to do if production is broken?
+Don't panic and follow the instructions below:
```
ncloudioj, line 42: Yep, don't panic!™ 😎

docs/dev/release-process.md
```diff
@@ -28,5 +28,27 @@ While the `[do not deploy]` can be anywhere in the title, it is recommended to p
...
+
+- depending on the severity of the problem, decide if this warrants [kicking off an incident][incident_docs];
+- if the root cause of the problem can be identified in a relatively small time, create a PR for the fix.
+    - verify the fix locally;
+    - verify the fix on `stage`, after it is reviewed by a Merino developer and merged;
+    - [deploy it to production](#releasing-to-production).
+
+**OR**
+
+- if the root cause of the problem is harder to track down, revert the last commit and then [deploy the revert commit to production](#releasing-to-production).
```
ncloudioj, line 52: Another clarification question: what if the problematic commit is not the last commit and some subsequent commits after it are also deployed in production. I think we'd want to provide developers with a way to easily revert to the last working commit. This can be tricky though with a highly active repo, but still worth thinking about.

ncloudioj (changes requested):
The changes for CircleCI look good. I left a few comments for the documentation. I know those comments/asks are not easy to fully expand in here, so could you please provide a little more clarifications here, we can tackle those remaining details in a follow up perhaps in a different format (such as a CD runbook on mana). Thanks!

docs/dev/release-process.md
Dexterp37: I was intentionally generic in this statement, because anyone could trigger, as long as they have write access to the Merino repository.

Ideally speaking, I would expect the individual who authored the PR to do that, but we don't strictly enforce this. The reason why I would expect that is that authors know the code and know what to watch out for when things are hitting production. If the author touched search providers in a subtle way and the plots show anomalies with search queries, then they might be better suited to identify the problem.

In general, we want developers / authors to feel the _responsibility_ of deploying to production, own that deployment and shepherd it to conclusion.

docs/dev/release-process.md
```diff
@@ -28,5 +28,27 @@ While the `[do not deploy]` can be anywhere in the title, it is recommended to p
...
+3. once in the "on hold" status, with all the other jobs successfully completed, clicking on the "thumbs up" action on the `unhold-to-deploy-to-prod` job row will approve it and trigger the deployment, unblocking the `deploy-to-prod` job;
+4. developers **must** monitor the [Merino Application & Infrastructure][merino_app_info] dashboard for any anomaly, for example significant changes in HTTP response codes, increase in latency, cpu/memory usage (most things under the infrastructure heading).
+
+[circleci_dashboard]: https://app.circleci.com/pipelines/github/mozilla-services/merino?branch=main&filter=all
+[merino_app_info]: https://earthangel-b40313e5.influxcloud.net/d/Cm83vS57z/merino-application-and-infrastructure?orgId=1&refresh=1m
+
+## What to do if production is broken?
+Don't panic and follow the instructions below:
+
+- depending on the severity of the problem, decide if this warrants [kicking off an incident][incident_docs];
```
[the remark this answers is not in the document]
Dexterp37: Yes, I believe this part should be changed / extended by somebody else rather than me. I meant to add this as an initial point of discussion, in order to have _something_. What I wrote is mostly high level and always relevant, regardless of the specifics of the incident and it boils down to this:

- if the problem is trivial and the fix is easy, create a PR, fix the thing and deploy;
- if the problem warrants a more thorough investigation, revert/turn off (if only we had feature flags) the latest change and investigate.

I'd recommend either leaving this text as it is or extending it in a separate PR as you see fit. I believe there's still value in slightly documenting the process to act on things when prod breaks.

docs/dev/release-process.md
Dexterp37: This is a great question, and I believe that this question boils down to the philosophical difference between Continuous Deployment and Continuous Delivery: with Continuous Delivery, you delivery whenever you feel like, maybe weekly, maybe not, but you potentially put together multiple changesets.

With Continuous Deployment, the idea is to deploy every time you merge to main. That's the part that reduces the risk of breaking, because when releasing on each single merge you can easily tell if you're merging something problematic. Ideally, you don't uncover with a deployment from today something that broke 15 days ago. Because the 15 days ago thing would have triggered problems on that deployment. If the interaction with the new thing is the problem, then you still solve it by reverting/disabling the newest change.

The need to go back in time to deploy an old commit, I believe, sort of defeats the point of Continuous Deployment. It's still doable, manually, by SRE though (which is the current state of things anyway).

.circleci/config.yml
```diff
@@ -49,13 +49,29 @@ workflows:
...
           requires:
             - checks
             - build-and-test
             - docker-image-build
             - contract-tests
+      # The following job will require manual approval in the CircleCI web application.
+      # Once provided, and when all the requirements are fullfilled (e.g. tests)
+      - unhold-to-deploy-to-prod:
+          <<: *main-filters
+          type: approval
```
hackebrot, line 63: TIL! 🤩

docs/dev/release-process.md
hackebrot: There's a lot of helpful context in your response, @Dexterp37. I would love this to be captured in the documentation.

.circleci/config.yml
```diff
@@ -181,8 +197,36 @@ jobs:
...
+      - skip-if-do-not-deploy
+      - setup_remote_docker
+      - attach_workspace:
+          at: workspace
+      - dockerhub-login
+      - run:
+          name: Load Docker image from workspace
+          command: docker load -i workspace/merino.tar.gz
+      - run:
+          name: Push to Dockerhub (prod)
```
hackebrot, line 226:
```suggestion
          name: Push to Docker Hub (prod)
```

docs/dev/release-process.md
```diff
@@ -28,5 +28,27 @@ While the `[do not deploy]` can be anywhere in the title, it is recommended to p
 The deployment pipeline will analyse the message of the merge commit (which will be contain the PR title) and make a decision based on it.
 
 ## Releasing to production
-The process to promote a build from `stage` to `production` is currently manually initiated by SRE.
-[This ticket](https://mozilla-hub.atlassian.net/browse/CONSVC-1566) (requires LDAP) deals with automating the process.
+Developers with write access to the Merino repository can initiate a deployment to production after a Pull-Request on the Merino GitHub repository is merged to the `main` branch. This can be done by:
+
+1. opening the [circleci dashboard][circleci_dashboard];
```
hackebrot, line 33:
```suggestion
1. opening the [CircleCI dashboard][circleci_dashboard];
```

docs/dev/release-process.md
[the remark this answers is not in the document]
hackebrot: I may be helpful to indicate that this section is subject to change.

hackebrot (commented):
Thank you for your work on this, @Dexterp37! 🚀

docs/dev/release-process.md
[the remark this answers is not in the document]
Dexterp37: I'd rather not mark content as subject to change for now. I believe the guidelines would be valid even if the context around them would change, but I'd defer this to future PRs to settle. (or really leave it to @ncloudioj )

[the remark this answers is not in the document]
ncloudioj: > I'd recommend either leaving this text as it is or extending it in a separate PR as you see fit. I believe there's still value in slightly documenting the process to act on things when prod breaks.

Sounds good to me. I think we will further expand this section with detailed guidance in a follow-up.

ncloudioj (approved):
@Dexterp37 Thanks for putting this up together!

Let's give it a spin. 🚀

Dexterp37:
> @Dexterp37 Thanks for putting this up together!
> 
> Let's give it a spin. 🚀

Thank you for the review :-)

I'll merge this even if the ci/checks fail, since they are not related to this PR. Note that this still requires a bit of work from Dustin on the Jenkins side to plug things together. He's doing the work today.

PR #374, opened 2022-04-26
chore: Remove merino-shepherd to favor consvc-shepherd

ncloudioj:
This is no longer needed as now we have https://github.com/mozilla-services/consvc-shepherd/.

cc: @tiftran


pjenvey (approved):
The latest clippy has a couple suggestions

merino-cache/src/redis/mod.rs
```diff
@@ -79,6 +79,7 @@ impl SimpleRedisLock {
     /// See if a record update is locked for pending update.
     ///
     /// This does not check lock value, only if a lock exists.
+    #[allow(clippy::wrong_self_convention)]
```
ncloudioj, line 82: We could rename this to something like `can_lock()`, but it really doesn't sound better than `is_locked()`. So I've decided to silence this warning.

ncloudioj: What do you think? @pjenvey

pjenvey: Definitely silence it, it's a bug they've already fixed: https://github.com/rust-lang/rust-clippy/pull/8738 :)

merino-adm/src/remote_settings/mod.rs
```diff
@@ -296,6 +296,11 @@ impl RemoteSettingsSuggester {
 
                     let icon_key = format!("icon-{}", adm_suggestion.icon);
                     let icon_url = if let Some(u) = icon_urls.get(&icon_key) {
+                        // Suppress the warning as `u` does not live long enough
+                        // since `from_maybe_shared()` expects its argument to
+                        // be either a reference type with static lifetime or an
+                        // owned type.
+                        #[allow(clippy::unnecessary_to_owned)]
                         Uri::from_maybe_shared(u.to_string()).expect("invalid URL")
```
pjenvey, line 304: I wonder if clippy's also confused about this one but I can't blame it, `from_maybe_shared` is a little confusing here. We can simplify it to `from_str(u)` w/ `use std::str::FromStr;` (or `try_from` which acts similarly but `from_str` reads better IMO) to avoid the clippy warning entirely.

```suggestion
                        Uri::from_str(u).expect("invalid URL")
```

pjenvey (changes requested):

merino-adm/src/remote_settings/mod.rs
ncloudioj: Ah, good point. Thanks!

pjenvey (approved):

PR #387, opened 2022-06-09
chore: Update sentry to the latest

ncloudioj:
Lots of security alerts with the old sentry clients, I guess we're now good to use the latest one.


merino-web/Cargo.toml
```diff
@@ -30,7 +29,7 @@ tracing = { version = "0.1.29", features = ["async-await"] }
 tracing-actix-web-mozlog = "0.4.1"
 tracing-futures = "0.2"
 tracing-actix-web = "0.5.0-beta.6"
-uuid = "0.8.2"
+uuid = { version = "0.8.2", features = ["serde"] }
```
ncloudioj, line 32: Somehow, Cargo won't build without this feature, had to explicitly enable it now.

pjenvey (approved):

PR #392, opened 2022-06-17
feat: Make Merino load suggestions from Enhanced Exposure

ncloudioj:
This fixes [CONSVC-1817](https://mozilla-hub.atlassian.net/browse/CONSVC-1817).

Change highlights:
* Merino now loads `offline-expansion-data` type suggestions from RemoteSettings if they're present
* The existing `data` type suggestions are used as the fallback if the former is absent. We can't merge those two sources in Merino due to business concerns. Eventually, suggestions with `offline-expansion-data` will replace those of `data` once Enhanced Exposure is ready to roll out to the entire population 
* Both `click_url` and `impression_url` fields are optional now. Those two fields are only used for suggestions that require external data sharing (e.g. sponsored suggestions). Note that Merino's API doc already notes that, but it was not reflected in the code

Notes:
* Due to data increase on Remote Settings, the suggestion downloading/loading time in Merino takes much longer than before (takes about 10 seconds on my machine using the production dataset), which significantly increased the startup delay of Merino. We might optimize this by using parallel loading
* The contract test is kinda hacky now as it currently doesn't support using different Kinto fixtures for different test scenarios, I ended up duplicating the fixture data to ensure all existing tests still work. We have filed [CONSVC-1821](https://mozilla-hub.atlassian.net/browse/CONSVC-1821) to improve it
* Our data model for suggestions is not flexible enough to cover all the use cases. Specifically, both `click_url` and `impression_url` are required if that suggestion provider expects us to share impression/click data with them, otherwise, it's fine to ignore them in the suggestion payload. E.g. we don't provide those URLs for the in-house Wikipedia suggestions. We might want to expand the existing data model to support those use cases better in the future


ncloudioj:
Hey @hackebrot @Trinaa, could you also take a look at this? The changes for the contract test can be found in the second commit. TIA!

test-engineering/contract-tests/kinto-setup/main.py
```diff
@@ -35,6 +35,9 @@ def main(
                 mimetype="application/json",
                 filecontent=data_file.read_bytes(),
             ),
+            data_type="offline-expansion-data"
+            if "offline" in data_file.stem
+            else "data",
```
Trinaa, line 40: I had a thought that an assignment with a common function, that assumes file name convention, but **removes the custom condition** would be more ideal for test. It also allows for more data types should the need arise.

Something like:

```suggestion
            data_type=data_file.stem[:-3],
```
```suggestion
            data_type=data_file.stem.rsplit('-', 1)[0],
```
```suggestion
            # Needs 'import re'
            data_type=re.match("^(?P<data_type>.*)-\d\d$", data_file.stem).group("data_type"),
```

Trinaa (commented):
A typo fix and a suggestion.
_\* This is my first CR at Mozilla, feedback on etiquette or process are welcome and appreciated_

merino-adm/src/remote_settings/mod.rs
```diff
@@ -241,8 +241,26 @@ impl RemoteSettingsSuggester {
...
-        data_records.retain(|r| r.get("type") == Some(&serde_json::json!("data")));
+        let all_data_records = records.clone();
+
+        // Get records whose `type` field is either `data` or `offline-expansion-data`,
+        // prefer `offline-expansion-data` over `data` if the former is present.
+        //
+        // See details: https://mozilla-hub.atlassian.net/browse/CONSVC-1817
+        let (mut data_records, offline_data_records): (Vec<_>, Vec<_>) = all_data_records
+            .into_iter()
+            .filter(|r| match r.get("type") {
```
pjenvey, line 252: could also use `matches!(r.get("type"), Some(...`

pjenvey (approved):

pjenvey (commented):
looks good

merino-adm/src/remote_settings/mod.rs
```diff
@@ -241,12 +250,28 @@ impl RemoteSettingsSuggester {
...
+        // Get records whose `type` field is either `data` or `offline-expansion-data`,
+        // prefer `offline-expansion-data` over `data` if the former is present.
+        //
+        // See details: https://mozilla-hub.atlassian.net/browse/CONSVC-1817
+        let (mut data_records, offline_data_records): (Vec<_>, Vec<_>) = all_data_records
+            .into_iter()
+            .filter(|r| {
+                matches!(
+                    r.get("type"),
+                    Some(&serde_json::Value::String(ref val)) if val == "data" || val =="offline-expansion-data"
```
pjenvey, line 264: whitespace nit (cargo fmt misses it since it's a macro)
```suggestion
                    Some(&serde_json::Value::String(ref val)) if val == "data" || val == "offline-expansion-data"
```

pjenvey (approved):

merino-adm/src/remote_settings/mod.rs
ncloudioj: oops 👍

ncloudioj:
Thank y'all for the review!

Will merge it once [this](https://github.com/mozilla-services/quicksuggest-rs/pull/23) gets merged first.

PR #394, opened 2022-06-22
chore: Skip circleci approval task for do-not-deploy tags [do not deploy]

ncloudioj:
This fixes [CONSVC-1779](https://mozilla-hub.atlassian.net/browse/CONSVC-1779).

Turns out that `circle-agent step halt` works at the job level rather than at the workflow level, that's why CircleCI still processes the subsequent jobs. Per this [thread](https://discuss.circleci.com/t/does-circleci-step-halt-works-with-version-2-1/36674/2), we have to cancel the entire workflow which I think it's reasonable in our case.

cc: @hackebrot

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -327,7 +327,9 @@ commands:
                 echo "Skipping remaining steps in this job: deployment was disabled for this commit."
                 circleci-agent step halt
 
-                # No need to deploy, just cancel the rest of jobs of the workflow
+                # No need to deploy, just cancel the rest of jobs of the workflow.
+                # See API detail: https://circleci.com/docs/api/v2/index.html#operation/cancelWorkflow
+
                 curl -X POST https://circleci.com/api/v2/workflow/${CIRCLE_WORKFLOW_ID}/cancel \
                 -H 'Accept: application/json' \
                 -H 'Circle-Token: ${SKIP_DEPLOY_API_TOKEN}'
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,7 +67,7 @@ workflows:
             - docker-image-build
             - contract-tests
       # On approval of the `unhold-to-deploy-to-prod` job, any successive job that requires it
-      # will run. In this case, it's manually triggering deployment to production. 
+      # will run. In this case, it's manually triggering deployment to production.
       - docker-image-publish-prod:
           <<: *main-filters
           requires:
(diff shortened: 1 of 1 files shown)


pjenvey (approved):

.circleci/config.yml
```diff
@@ -326,4 +326,9 @@ commands:
             if git log -1 "$CIRCLE_SHA1" | grep -q '\[do not deploy\]'; then
                 echo "Skipping remaining steps in this job: deployment was disabled for this commit."
                 circleci-agent step halt
+
+                # No need to deploy, just cancel the rest of jobs of the workflow
```
hackebrot, line 330: Do you mind adding a link to documentation for this API endpoint? Does it cancel all jobs in the workflow right away or does it wait for running jobs to complete?

.circleci/config.yml
```diff
@@ -326,4 +326,9 @@ commands:
             if git log -1 "$CIRCLE_SHA1" | grep -q '\[do not deploy\]'; then
                 echo "Skipping remaining steps in this job: deployment was disabled for this commit."
                 circleci-agent step halt
+
```
hackebrot, line 329: Does it make sense to keep the `circleci-agent step halt` call here or should we remove that?

hackebrot (commented):
Hi @ncloudioj! 👋🏻 

Thank you for picking this up. I've added two questions inline.

.circleci/config.yml
ncloudioj: Sure, will do.

I couldn't find details in the API doc about how the cancellation is done. In this case, both `docker-image-publish-stage` and `docker-image-publish-prod` requires other jobs to be run first, and we cancel the workflow within those two jobs, so I think there should be no other jobs running at that point.

In fact, the `skip-if-do-not-deploy` command should cancel the workflow when `docker-image-publish-stage` gets run.

hackebrot (commented):
LGTM! 🚀

.circleci/config.yml
ncloudioj: No worries. Hmm, looks like we will need a follow-up for this. It still asked me to approve when I merged this PR to main. Investigating.

hackebrot: Hi @ncloudioj! 👋🏻 

Did you have a chance to look into this?

ncloudioj: @hackebrot Just wanted to circle back on this, turns out that CircleCI was also not happy with using a personal API token here, a "permission denied" error was thrown when trying to cancel the rest of build tasks on CircleCI. Looks like we need to set up another API token just CI for this. Will reach out to SRE folks for some help.

hackebrot: Thanks, @ncloudioj! I've been following your conversation with @dlactin on Slack. 👍🏻

PR #408, opened 2022-08-09
added request delay field with default and implementation in test_merino [do not deploy]

taddes:
WIP

I followed the spec in the Pydantic docs [here](https://pydantic-docs.helpmanual.io/usage/models/#required-fields) and while there are some other ways to go about handing these optional fields, it can be can discussed them further based on your thoughts @Trinaa. I see the use of `default_factory` a few times which we could make use of given the hope of breaking up Merino and Kinto requests.  

I also saw something like this could work, however we don't want it to occur for each run of the test, but I wanted to point it out as an option nonetheless:
`@pytest.fixture(autouse=True)
def slow_down_tests():
    yield
    time.sleep(1)`

Post discussion and walking through both changes in our separate branches, we can move out of WIP status.

### DIFF
--- a/test-engineering/contract-tests/client/tests/models.py
+++ b/test-engineering/contract-tests/client/tests/models.py
@@ -22,7 +22,7 @@ class Request(BaseModel):
     path: str
     headers: List[Header] = []
     # Delay is optional, providing time for data refresh
-    delay: Optional[int] = 0
+    delay: Optional[float] = 0
 
 
 class Suggestion(BaseModel, extra=Extra.allow):
--- a/test-engineering/contract-tests/client/tests/models.py
+++ b/test-engineering/contract-tests/client/tests/models.py
@@ -22,7 +22,7 @@ class Request(BaseModel):
     path: str
     headers: List[Header] = []
     # Delay is optional, providing time for data refresh
-    delay: Optional[float] = 0
+    delay: Optional[float] = None
 
 
 class Suggestion(BaseModel, extra=Extra.allow):
--- a/test-engineering/contract-tests/client/tests/models.py
+++ b/test-engineering/contract-tests/client/tests/models.py
@@ -21,6 +21,8 @@ class Request(BaseModel):
     method: str
     path: str
     headers: List[Header] = []
+    # Delay is optional, providing time for data refresh
+    delay: Optional[int] = 0
 
 
 class Suggestion(BaseModel, extra=Extra.allow):
--- a/test-engineering/contract-tests/client/tests/test_merino.py
+++ b/test-engineering/contract-tests/client/tests/test_merino.py
@@ -6,6 +6,7 @@
 from typing import Dict, List, Set, Tuple
 
 import pytest
+import time
 import requests
 from models import ResponseContent, Step, Suggestion
 
@@ -69,7 +70,8 @@ def assert_200_response(
 
         if "wiki_fruit" in suggestion.provider:
             # The icon URL is static for WikiFruit suggestions
-            expected_suggestion = expected_suggestions_by_id[suggestion_id(suggestion)]
+            expected_suggestion = expected_suggestions_by_id[suggestion_id(
+                suggestion)]
             assert suggestion.icon == expected_suggestion.icon
             continue
 
@@ -84,6 +86,11 @@ def test_merino(merino_url: str, steps: List[Step], kinto_icon_urls: Dict[str, s
         method = step.request.method
         url = f"{merino_url}{step.request.path}"
         headers = {header.name: header.value for header in step.request.headers}
+        delay = step.request.delay
+
+        # Process delay if defined in request model
+        if delay > 0:
+            time.sleep(delay)
 
         r = requests.request(method, url, headers=headers)
 
--- a/test-engineering/contract-tests/client/tests/test_merino.py
+++ b/test-engineering/contract-tests/client/tests/test_merino.py
@@ -89,7 +89,7 @@ def test_merino(merino_url: str, steps: List[Step], kinto_icon_urls: Dict[str, s
         delay = step.request.delay
 
         # Process delay if defined in request model
-        if delay > 0:
+        if (delay := step.request.delay) is not None:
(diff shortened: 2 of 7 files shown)


test-engineering/contract-tests/client/tests/test_merino.py
```diff
@@ -69,7 +70,8 @@ def assert_200_response(
 
         if "wiki_fruit" in suggestion.provider:
             # The icon URL is static for WikiFruit suggestions
-            expected_suggestion = expected_suggestions_by_id[suggestion_id(suggestion)]
+            expected_suggestion = expected_suggestions_by_id[suggestion_id(
+                suggestion)]
```
Trinaa, line 74: It's not documented, but we are using 'black' to format. Perhaps give that a run, since I think it will impact this change.

test-engineering/contract-tests/client/tests/test_merino.py
```diff
@@ -6,6 +6,7 @@
 from typing import Dict, List, Set, Tuple
 
 import pytest
+import time
```
Trinaa, line 9: Suggest to group the std library imports.

```suggestion
import time
from typing import Dict, List, Set, Tuple

import pytest
```

test-engineering/contract-tests/client/tests/models.py
```diff
@@ -21,6 +21,8 @@ class Request(BaseModel):
     method: str
     path: str
     headers: List[Header] = []
+    # Delay is optional, providing time for data refresh
+    delay: Optional[int] = 0
```
Trinaa, line 25: Should this be a float?

I just saw that the 'sleep' function supports it, so perhaps we should propagate the precision?
image

Trinaa (commented):
Hi @taddes 👋🏻 

I vote for the solution you implemented over the other options (`default_factory` and `pytest.fixture`) It's easy to understand and use.

I noted some other observations for your consideration, but nothing critical jumps out at me.

I think we should document this 'delay' feature in the client [README.md](https://github.com/mozilla-services/merino/blob/main/test-engineering/contract-tests/client/README.md), but perhaps this should be part of a larger documentation task update (thoughts?). I think a `steps` section, like the one in [contile](https://github.com/mozilla-services/contile/blob/main/test-engineering/contract-tests/client/README.md) might be called for.

test-engineering/contract-tests/client/tests/models.py
taddes: Good point, will change to a float!

test-engineering/contract-tests/client/tests/test_merino.py
taddes: Good point, done!

taddes:
Hi @Trinaa , thanks for making those observations. I took care of those small reformat and thanks for pointing out the [do not deploy] annotation in the PR titles.

I agree that actually expanding the documentation to be more in line like the contile docs is a good idea, as the README.md is quite spartan.  Perhaps we make a task for that, though we can at least make note of the changes and incrementally expand it.

Trinaa (approved):

Trinaa:
@taddes just noticed there is no `test:` prefix on the commit, might want to add that before merge.

test-engineering/contract-tests/client/tests/models.py
```diff
@@ -21,6 +21,8 @@ class Request(BaseModel):
     method: str
     path: str
     headers: List[Header] = []
+    # Delay is optional, providing time for data refresh
+    delay: Optional[float] = 0
```
ncloudioj, line 25: Normally, optional fields will use `None` as the default value. Did you want this field to be a float with 0 as the default instead?

taddes: @ncloudioj yes! If the `None` type is passed to `time.sleep()` the function will crash, so I wanted to make sure we strictly set it to 0. Given that optional fields use `None` as a default, do you see a problem with this implementation that overrides it?

ncloudioj: I think if delay is optional, you could do:

```
if (delay := step.request.delay) is not None:
    time.sleep(delay)
```

taddes: Goodness @ncloudioj ... the walrus operator in the wild! Isn't this getting fancy.

ncloudioj: Haha, many also dislike it. I think it's pretty handy.

ncloudioj (approved):
Looks good, thanks!

taddes:
Any chance I can get someone's eyes on this CircleCI error @ncloudioj  or @pjenvey ? Seen this before? It just came up between pushes, even though my previous approval passed... quite odd. Not sure if I really should add Eq to the derive and doesn't appear like my branch is behind.

From CircleCI
```
error: you are deriving `PartialEq` and can implement `Eq`
  --> merino-settings/src/logging.rs:71:24
   |
71 | #[derive(Debug, Clone, PartialEq)]
   |                        ^^^^^^^^^ help: consider deriving `Eq` as well: `PartialEq, Eq`
   |
   = note: `-D clippy::derive-partial-eq-without-eq` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq

error: could not compile `merino-settings` due to previous error
warning: build failed, waiting for other jobs to finish...

Exited with code exit status 101
```

ncloudioj:
Doh, looks like Clippy had some suggestions about the Rust code. @taddes Do you mind piggybacking that by following its suggestions?

taddes:
Yup! Adding the Eq trait as requested by Sir Clippy. Thanks for confirming it's not anything weird @ncloudioj .

pjenvey (approved):
Yup, new Rust release came out today and we're still using rust:latest

taddes:
Thx @pjenvey ! Still getting used to that. What seems like magical Rust unicorns dropping rainbows on our code.

PR #417, opened 2022-09-12
test: add contract tests for remote settings refresh [do not deploy]

Trinaa:
## Description
* add test scenario for remote settings refresh

⚠️ **Note**: These changes should be identical to https://github.com/mozilla-services/merino-py/pull/48

## Issue(s)
[CONSVC-1866](https://mozilla-hub.atlassian.net/browse/CONSVC-1866)

### DIFF
--- a/config/providers/ci.yaml
+++ b/config/providers/ci.yaml
@@ -2,9 +2,7 @@
 test_wiki_fruit:
   type: wiki_fruit
 adm:
-  type: memory_cache
-  inner:
-    type: remote_settings
-    collection: quicksuggest
-    suggestion_score: 0.3
-    resync_interval_sec: 2
+  type: remote_settings
+  collection: quicksuggest
+  suggestion_score: 0.3
+  resync_interval_sec: 2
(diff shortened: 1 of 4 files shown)


Trinaa:
@ncloudioj the test 'remote_settings__refresh' is passing with merino-py ATM and I'm surprised to see it failing for merino. I want to double check that expectation and wonder if there is a major difference between merino and merino-py that @taddes and I are unaware of? Perhaps with the local cache?

From the [logs](https://github.com/mozilla-services/merino-py/pull/48), I can see that merino has an opportunity to re-sync with RS before the test makes it's request for 'tree', but I also see this triggering a 'cache hit' in merino. Am I reading this right? Is the cache interfering? 
image

ncloudioj:
Oh, looks like the Remote Settings provider is wrapped inside of a memory cache provider (see [here](https://github.com/mozilla-services/merino/blob/main/config/providers/ci.yaml#L5)) in the `ci` environment, that was why Merino fetched from the cache rather than Remote Settings in this case.

Hmm, I was under the impression that we removed this memory cache wrapper earlier in another patch, perhaps that was for Merino-py? Anyway, dropping the wrapper should fix this.

ncloudioj (approved):

